PR #12, opened 2018-09-30
first iteration team datamodel Closes #4

### LINKED ISSUE #4: Create data model for teams
In `model/team.py`, create a class Team that we'll use to represent teams. At the moment in Rocket teams have:

* Display name
* Platform
* List of members
* GitHub team name

In particular we will need to decide how to represent the list of members internally - potentially it could be a list of User objects (see #3), some kind of identifier that could be referenced externally, or something else.

rmcreyes:
Created data model for a team with generic getters and setters.


vyzw:
A dict instead of a list might make more sense if there's no concept of order in a team. And I'm thinking of using a property called slackID (which was in rocket 1) which I assume lets rocket know which slack user we're referring to instead of uuid.

rmcreyes:
> A dict instead of a list might make more sense if there's no concept of order in a team. And I'm thinking of using a property called slackID (which was in rocket 1) which I assume lets rocket know which slack user we're referring to instead of uuid.

Ok sure, what would the mapping be like for a dict?

rwblickhan:
I think a set would be better than a dict - we're not mapping the user IDs to anything, we just want to iterate over the users and check if a user is contained.

rmcreyes:
> I think a set would be better than a dict - we're not mapping the user IDs to anything, we just want to iterate over the users and check if a user is contained.

Yeah that makes sense, I agree.

PR #16, opened 2018-10-02
Setup containerization

### LINKED ISSUE #6: Setup containerization
Rocket 2.0 should be containerized and this needs to be set up.

cheukyin699:
Closes #6.

Near-empty Dockerfile added. Since there isn't much going on, only setup stuff is included (python image and ruby gem installs).

**Question**: Is there an alternative to the ruby gem `mdl`? Specifically, a python package? It would make the image a bit smaller (right now we are installing the ruby/rubygems package for the gem).


rwblickhan:
> Is there an alternative to the ruby gem mdl? Specifically, a python package?

Hadn't thought about that. A quick Google search came up with [`pymarkdownlint`](https://github.com/jorisroovers/pymarkdownlint), which is a clone of `mdl` in Python, though it's now inactive. If you'd like to add try it out and add it to this pull request, you can go ahead - I assume it still works fine even though it's not under active development. If you do add it, make sure to update the Dockerfile and also .travis.yml.

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+# Set up working directory
+WORKDIR /app
+
+# Copy everything into working directory
+COPY . /app
+
+# Install everything
+RUN apt-get update && apt-get install -y rubygems
+RUN pip install pipenv
+RUN pipenv install --dev
```
rwblickhan: This can just be `pipenv install` - the container is primarily for deployment, so we can probably drop the development dependencies.

cheukyin699:
> If you'd like to add try it out and add it to this pull request, you can go ahead - I assume it still works fine even though it's not under active development.

Got an import error with `pymarkdownlint`. Checked it and found that the project was using python 2.7 `ConfigParser`. I think we will stick with Ruby `mdl` - I don't think I want to maintain this project after ~3 years of inactivity.

rwblickhan:
> I think we will stick with Ruby `mdl` - I don't think I want to maintain this project after ~3 years of inactivity.

Haha fair enough - though I was going to say, if we run out of features... 😜

PR #21, opened 2018-10-03
Add User class and properties.

### LINKED ISSUE #3: Create data model for users
In `model/user.py`, create a class User that we'll use to represent users. At the moment in Rocket users have:
* Name
* Email
* Position
* GitHub username
* Major
* Bio

In the future we will also need to add permissions in the future and potentially other fields, so it should be somewhat extensible.

vyzw:
# Pull Request

## Description

Give a brief description of your changes: Added the User class, as well as its properties.

## Testing

## Ticket(s)

Closes #3 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .idea/
+.vscode/
--- a/model/user.py
+++ b/model/user.py
@@ -1,8 +1,10 @@
 """Data model to represent an individual user."""
-from permissions import Permissions
+from model.permissions import Permissions
 
 
 class User:
+    """Represent a user with related fields and methods."""
+
     __slack_id = ""
     __name = ""
     __email = ""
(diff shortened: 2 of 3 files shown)


rwblickhan:
It looks like the Travis build is failing because it's missing [docstrings](https://www.python.org/dev/peps/pep-0257/) - can you please add those? The issues are listed in the build log here: https://travis-ci.org/ubclaunchpad/rocket2.0/builds/436471097. You can also run `pydocstyle` locally and get the same output, though if this branch is out of date you probably don't have it as a dependency yet - if you want it, you could try updating your branch as described at the bottom of [CONTRIBUTING](https://github.com/ubclaunchpad/rocket2.0/blob/master/CONTRIBUTING.md).

.gitignore
```diff
@@ -1 +1,2 @@
 .idea/
+.vscode/
```
rwblickhan: 👍

model/user.py
```diff
@@ -1 +1,62 @@
 """Data model to represent an individual user."""
+from permissions import Permissions
```
rwblickhan: Looks good to me. Can you write unit tests for all these methods? (I forgot to include it in the original ticket. 😞) You can probably follow Robin's [teams tests](https://github.com/ubclaunchpad/rocket2.0/blob/master/tests/model/team_test.py) pretty closely.

rwblickhan: Oh, also, do keep in mind that any tests you add will also need docstrings.

model/user.py
```diff
@@ -1 +1,62 @@
...
+
+
+class User:
```
cheukyin699: @rwblickhan Does the user need a uuid as well?

rwblickhan: I’d say just using Slack ID is probably good enough, since those should be unique and we want to uniquely tie our users to Slack users. But I’m open to other ideas as well!

PR #35, opened 2018-10-06
Add pre-commit/post-merge hooks

### LINKED ISSUE #15: Create pre-commit hook
Travis is protecting our main branches but it’s not officially documented what our build is doing, nor is it very convenient to run all those commands locally. So, we should add a build script to run all those commands and have contributors add this as a pre-commit hook, so that we don’t have to run a Travis build to find linting errors.

rwblickhan:
# Pull Request

## Description

Adds pre-commit and post-merge hooks and a Makefile to install them, so that dependencies are kept up to date and we're less likely to commit something that will fail the Travis build.

## Ticket(s)

Closes #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,3 +33,11 @@ pipenv shell
 ```
 
 and then commands like `pycodestyle` and `pytest` can be run like normal.
+
+Additionally, we use [Travis CI](https://travis-ci.org/ubclaunchpad/rocket2.0) as
+a CI system. To run the same checks locally, please install our git hooks:
+
+```bash
+cd scripts/
+make install
+```
--- a/README.md
+++ b/README.md
@@ -35,7 +35,15 @@ pipenv shell
 and then commands like `pycodestyle` and `pytest` can be run like normal.
 
 Additionally, we use [Travis CI](https://travis-ci.org/ubclaunchpad/rocket2.0) as
-a CI system. To run the same checks locally, please install our git hooks:
+a CI system. To run the same checks locally, we provide `scripts/build_check.sh`;
+this can be run with:
+
+```bash
+./scripts/build_check.sh
+```
+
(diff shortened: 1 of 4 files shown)


scripts/build_check.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+pipenv run pycodestyle .
+pipenv run pydocstyle .
+mdl .
```
cheukyin699: Maybe switch to using `bundle exec mdl` with `Gemfile`? Though this raises the complexity a tad bit (with the extra 1-line-long file).

Or we could remove it entirely and switch to the unmaintained python one.

I personally like the latter.

README.md
```diff
@@ -32,4 +32,12 @@ environment like so:
 pipenv shell
 ```
 
-and then commands like `pycodestyle` and `pytest` can be run like normal.
\ No newline at end of file
+and then commands like `pycodestyle` and `pytest` can be run like normal.
+
+Additionally, we use [Travis CI](https://travis-ci.org/ubclaunchpad/rocket2.0) as
+a CI system. To run the same checks locally, please install our git hooks:
```
cheukyin699: In addition to installer git hooks, can also mention that user/dev can run the scripts in the folder to achieve the same effect, without committing/merging.

scripts/build_check.sh
cheukyin699: Never mind. I think it's just that my ruby/gem installation is busted. Resolving.

rwblickhan: Hmm this does seem problematic. I'm getting issues with `gem install ...` as well (I have to use sudo) but then installing bundler requires... `gem install bundler`. Perhaps we should just come back to this later?

Issue #27, opened 2018-10-06
Implement Slack client "send message" methods


rwblickhan:
Design and implement methods in the Slack client class from #26 so that other modules can send direct messages to a user or send a message to a channel.

Issue #7, opened 2018-09-29
Create Slack client


rwblickhan:
In `bot/bot.py` (or a similar file) create a new Slack client per the [Python Slack library](https://github.com/slackapi/python-slackclient). In particular it should connect to the Slack bot created in #1 - for now this can be hardcoded but eventually we will want it to be configurable.

PR #41, opened 2018-10-13
create bot class

michaelenglo:
# Pull Request

## Description
Implement Bot class to interface with Slack

## Testing

## Ticket(s)

Affects #7 #27 


Pipfile
```diff
@@ -5,12 +5,14 @@ name = "pypi"
...
 [packages]
 pytest = "*"
+slackclient = "*"
 
 [dev-packages]
 pycodestyle = "*"
 codecov = "*"
 pytest-cov = "*"
 pydocstyle = "*"
+pylint = "*"
```
rwblickhan: Why did you add `pylint`?

bot/bot.py
```diff
@@ -0,0 +1,20 @@
+import os
```
rwblickhan: Please add [docstrings](https://www.python.org/dev/peps/pep-0257/) for this module and its functions. (This is causing the Travis build to fail at the moment.)

bot/bot.py
```diff
@@ -0,0 +1,20 @@
...
+from slackclient import SlackClient
+
+slack_token = os.environ["SLACK_BOT_TOKEN"]
+
+class Bot:
+
+  def __init__(self):
+    self.sc = SlackClient(slack_token)
+
+  def sendMsg(self, message, channel):
```
rwblickhan: Can you add unit tests for this method? You will probably want to use [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html) to mock the `SlackClient` (let me know if you want some help looking at that - it may require you to refactor the class a bit to follow [dependency inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle) better).

bot/bot.py
```diff
@@ -0,0 +1,40 @@
+"""Utility class for interacting with Slack API"""
+import os
+from slackclient import SlackClient
+
+slack_token = os.environ["SLACK_API_TOKEN"]
```
rwblickhan: It would probably be a good idea to move this into the `__init__` function for the actual bot class. (Also, we would want to make this configurable at some point, e.g. via a yaml config file, but that doesn't need to be in this pull request.)

bot/bot.py
```diff
@@ -0,0 +1,40 @@
...
+  def send_dm(self, message, slack_user_id):
+    """Send direct message to user with id of slack_user_id"""
+    response = self.sc.api_call(
+      "chat.postMessage",
+      channel=slack_user_id,
+      text=message
+    )
+
+    if not response['ok']:
+      raise Exception('API Call error: {}'.format(response["error"]))
```
rwblickhan: This is fine for now but I think in the long run raising exceptions will be somewhat problematic - once we set up logging we'll probably want this to log an error but otherwise return silently.

bot/bot.py
```diff
@@ -0,0 +1,40 @@
...
+    response = self.sc.api_call(
+      "chat.postMessage",
+      channel=channel_name,
+      text=message
+    )
+
+    if not response['ok']:
+      raise Exception('API Call error: {}'.format(response["error"]))
+
+    return
```
rwblickhan: You don't need an explicit `return` at the end of a method if you're not returning anything.

tests/bot/bot_test.py
```diff
@@ -0,0 +1,20 @@
+"""Test Bot Class."""
+from bot.bot import Bot
+
+def test_send_dm():
+    """Test the Bot class method send_dm()."""
+    bot = Bot()
```
rwblickhan: These unit tests should probably be using mocks, so that we're not sending messages every time we run the unit tests.

bot/bot.py
cheukyin699: Also try `os.environ.get('SLACK_API_TOKEN', <default value>)` to prevent key errors.

michaelenglo:
@rwblickhan I am not sure how to best resolve the Pipfile.lock conflict here. Any thoughts?

rwblickhan:
This branch is out of date now, right? It’s probably best to bring it up to date with a rebase, though that would also require fixing the `Pipfile.lock` conflicts. Alternatively we could do some “git surgery” with cherry-picks and all, though we’d want to be careful about that. Maybe it’s best to discuss it this weekend.

Issue #31, opened 2018-10-06
Stub out `user` command


rwblickhan:
Create a user command class in `command/commands/`, stub out its main methods (e.g. `handle()`), and implement  `help()` and `name()`.

rwblickhan:
Coordinate with @rmcreyes as he'll be responsible for creating the `command/` directory (in #29) and may end up naming it something different.

rwblickhan:
@bfbachmann also suggests we look into [abstract base classes (ABCs)](https://docs.python.org/2/library/abc.html) which are basically Python's version of interfaces. However, do note that they can't do as much as metaclasses, so those are still likely a good idea in this case.

PR #49, opened 2018-10-18
Added user command class with parsers and handler stubs

jacqueschen1:
# Pull Request

## Description

Added user command class with parsers using argparse and handler stubs. Test class also added.

## Testing

## Ticket(s)

Affects #31 

### DIFF
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -1,0 +1,89 @@
+"""Command parsing for user events"""
+import argparse
+import shlex
+
+
+class UserCommand:
+
+    __command_name = "user"
+    __help = "User Command Reference:\n\n @rocket user\n\n Options:\n\n edit \n --name NAME\n" \
+             " --email ADDRESS\n --pos YOURPOSITION\n --major YOURMAJOR\n --bio YOURBIO\n 'edit properties of your" \
+             "Launch Pad profile\n surround arguments with spaces with single quotes'" \
+             "\n ADMIN/TEAM LEAD ONLY option: --member MEMBER_ID\n 'edit properties of another" \
+             "user's Launch Pad profile'\n\n view MEMBER_ID\n 'view information about a user'\n\n " \
+             "help\n 'outputs options for user commands'\n\n " \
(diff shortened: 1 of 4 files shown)


command/commands/user.py
```diff
@@ -0,0 +1,89 @@
+"""Command parsing for user events"""
+import argparse
+import shlex
+
+
+class UserCommand:
+
+    __command_name = "user"
+    __help = "User Command Reference:\n\n @rocket user\n\n Options:\n\n edit \n --name NAME\n" \
```
rwblickhan: A small thing (that doesn't have to be changed in this PR, though we should make an issue): eventually we will probably want this to be put into some kind of config file (probably specific to the UserCommand module?) rather than have it hardcoded.

command/commands/user.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_name(self):
+        """Return the command type."""
+        return self.__command_name
+
+    def get_help(self):
+        """Return command options for user events."""
+        return self.__help
+
+    def handle(self, command):
```
rwblickhan: One thing that I forgot about is that we'll want to have some way to tell what user sent the command, for permissions purposes, so eventually this will have to change to accommodate that - I'm thinking that the easiest way to do that would be to add an extra parameter for "permissions level" and then Robin's "core framework" would be responsible for determining what the correct permissions level is (though then that begs the question of how the core framework can tell...).

command/commands/user.py
```diff
@@ -0,0 +1,89 @@
...
+        """Handle command by splitting into substrings and giving to parser."""
```
rwblickhan: It would be nice to have a `:param` line in the docstring.

command/commands/user.py
jacqueschen1: That's true, it's probably best to add that later though since we're not sure how that's gonna look like yet.

command/commands/user.py
jacqueschen1: Would it be a `:param` for every command? Would it also include all the options for user edit?

rwblickhan: I told you this over Slack but just so it’s public, it’s part of a docstring that lists the parameters for a function (one for each parameter).

PR #53, opened 2018-10-20
Create database facade

### LINKED ISSUE #30: Create database facade
Create a database facade (probably in a `database/` directory) that exposes core database actions that commands/events can interact with. For now, focus on user-related actions; these can just be stubbed out, since the point of this ticket is more to get the database facade's API up and going so other modules can start calling it.

This doesn't necessarily have to be a whole class; it might be better to leave it as a set of functions (but I leave that up to the developers!).

cheukyin699:
# Pull Request

## Description

Add database facade stubs (users only) and some basic tests.

## Ticket(s)

Closes #30


Pipfile
```diff
@@ -5,6 +5,7 @@ name = "pypi"
 
 [packages]
 pytest = "*"
+"boto3" = "*"
```
rwblickhan: Is there a reason why there's quotation marks here? We should remove them for consistency if they're not needed.

db/facade.py
```diff
@@ -0,0 +1,61 @@
+"""Database Facade."""
+# import boto3
```
rwblickhan: The DynamoDb specific code should be refactored into its own module. But we can keep it here for now so that we can get this pull request done and start working calling it in other modules.

cheukyin699: Gotcha. That's the next thing to work on.

cheukyin699:
@rwblickhan I think the requirements are a bit steep (100% coverage).

rwblickhan:
Hmm yeah, that should’ve fixed in my last PR but I guess the “patch” check is controlled by a different config (the Codecov configs are... not super well documented). I’ll put up another pull request tomorrow if I can figure out how the patch check is supposed to be configured (or I’ll just disable it if I can’t... project coverage is really what we care about).

rwblickhan:
Sorry for the delay, I've made the codecov/patch check optional, though annoyingly it still says "checks not successful." I'll look into turning it off entirely. In any case this should be ready to merge once it's reviewed again.

db/facade.py
```diff
@@ -0,0 +1,61 @@
...
+        TODO: Retrieve user from users table.
+
+        :return: returns a user model if slack id is found.
+        """
+        return User(slack_id)
+
+    def query_user(self, parameter):
+        """
+        TODO: Query for specific users by parameter.
+
```
rwblickhan: What will the `parameter` field look like? Admittedly there's a reason we're using Python (we can always decide later!) but this would be a good thing to think about.

db/facade.py
```diff
@@ -0,0 +1,67 @@
...
+        """Return a string representing this class."""
+        return "Database Facade"
+
+    def store_user(self, user):
+        """
+        Store user into users table.
+
+        :param user: A user model to store
+        """
+        # Assume that the tables are already set up this way
```
rwblickhan: Keep in mind that eventually we'll want some error handling to check this... that should probably be handled by the DynamoDb-specific portion of the code, though.

PR #57, opened 2018-10-27
Add __str__ to data model classes

### LINKED ISSUE #47: Add __str__ to data model classes
For the purposes of logging/debugging it would be useful to add a [`__str__()` special method](https://docs.python.org/3/reference/datamodel.html#object.__str__) for each data model class that prints out the data that it contains.

jacqueschen1:
# Pull Request

## Description

Added __str__ for team and user in model

## Testing

## Ticket(s)

Closes #47 


model/team.py
```diff
@@ -49,3 +49,11 @@ def get_members(self):
     def is_member(self, slack_id):
         """Identify if any member has the ID specified in the argument."""
         return slack_id in self.__members
+
+    def __self__(self):
+        """Prints information on the team class."""
+        return "Team, github name: {}, display name: {}. platform: {}" \
```
rwblickhan: A teensy tiny formatting thing: for this (and for user) I would personally prefer if it printed something like "Team - Github name: {}, display name: {}, ..." (that is, instead of a comma after "Team").

tests/command/commands/user_test.py
```diff
@@ -28,6 +28,12 @@ def test_get_help():
     assert testcommand.get_help() == help_text
 
 
+def test_handle_default():
+    """Test user command default @rocket user parser."""
+    testcommand = UserCommand()
+    assert testcommand.handle('user') == help_text
```
rwblickhan: 👍

tests/model/team_test.py
```diff
@@ -64,3 +64,9 @@ def test_is_member():
     assert team.is_member(new_slack_id) is False
     team.add_member(new_slack_id)
     assert team.is_member(new_slack_id)
+
+
+def test_print():
+    """Test print team class."""
+    team = Team("brussel-sprouts", "Brussel Sprouts")
+    print(team)
```
rwblickhan: A better test here might be to get the string from the Team and checking if its what you expect, rather than printing it (where you would have to check by hand that the unit test succeeded).

tests/model/user_test.py
```diff
@@ -85,3 +85,9 @@ def test_set_permissions_level():
     user = User("U0G9QF9C6")
     user.set_permissions_level(Permissions.admin)
     assert user.get_permissions_level() == Permissions.admin
+
+
+def test_print():
+    """Test print user class."""
+    user = User("U0G9QF9C6")
```
rwblickhan: Same.

model/user.py
```diff
@@ -76,3 +76,13 @@ def get_permissions_level(self):
     def set_permissions_level(self, permissions_level):
         """Set this user's permissions level to the given argument."""
         self.__permissions_level = permissions_level
+
+    def __str__(self):
+        """Prints information on the user class."""
+        return "User, slack_id: {}, name: {}, email: {}, github: {}," \
```
rwblickhan: Same.

rwblickhan:
It looks like the tests are failing. Also, we might want to do some git surgery to clean up the branch before merging (although we could squash-merge instead) - it looks like this branch picked up a bunch of old commits.

tests/model/team_test.py
jacqueschen1: How do you get the team string without printing it?

rwblickhan: Call `__str__()` directly or look into the formatting functions in Python.

model/team.py
cheukyin699: Or would something like `return str(self.__dict__)` be better because of the jsonish format? Just a thought.

tests/model/team_test.py
cheukyin699: Would like to add that calling `str(team)` also works.

model/team.py
jacqueschen1: True @rwblickhan would you prefer to use `return str(self.__dict__)` instead?

rwblickhan: Up to you two - that would definitely look a bit clearer code-wise.

Pipfile
```diff
@@ -6,6 +6,7 @@ name = "pypi"
 [packages]
 pytest = "*"
 boto3 = "*"
+pydocstyle = "*"
```
rwblickhan: This should be under `[dev-packages]`, no?

command/commands/user.py
```diff
@@ -6,49 +6,58 @@
...
-             " --major YOURMAJOR\n --bio YOURBIO\n" \
-             " 'edit properties of your Launch Pad profile\n" \
-             " surround arguments with spaces with single quotes'" \
-             "\n ADMIN/TEAM LEAD ONLY option: --member MEMBER_ID\n" \
-             " 'edit properties of another user's Launch Pad profile'\n\n" \
-             " view MEMBER_ID\n 'view information about a user'\n\n " \
-             "help\n 'outputs options for user commands'\n\n " \
-             "ADMIN ONLY\n\n delete MEMBER_ID\n" \
-             " 'permanently delete member's Launch Pad profile'"
+    def __init__(self):
```
rwblickhan: In the future it would be better to put these changes in a different pull request, so they can be reviewed independently (though since it's already here, don't bother changing it for this case).

model/team.py
jacqueschen1: I think it would look clearer

Pipfile
jacqueschen1: Uhhh I'm not sure why that's changed strange

command/commands/user.py
jacqueschen1: Sounds good

Pipfile
rwblickhan: git weirdness 🤷‍♀️ Probably a good idea to remove it though.

jacqueschen1:
I'm not sure why the Travis fail is failing. When I run pytest locally it passes, but there's also only 36 tests, as compared to the 40 tests when run on Travis.

rwblickhan:
It looks like the problem is that the order of the two users added to the team isn’t guaranteed in `test_print`, so it occasionally changes the order and the test fails. That seems like a bit of a problem - either that unit test should only add one user or perhaps our Python master @cheukyin699 has a recommendation?

cheukyin699:
2 ways to do this. Either we start using `unittest.TestCase` class (in which case we could use the `self.assertCountEqual`), or we just do 2 manual `assert` statements, particularly in a loop (check to see if each element in `l1` is also in `l2`).

For the first option, we'd have to refactor everything to use the classes for consistency.

command/commands/user.py
```diff
@@ -6,49 +6,58 @@
...
-             " --email ADDRESS\n --pos YOURPOSITION\n" \
-             " --major YOURMAJOR\n --bio YOURBIO\n" \
-             " 'edit properties of your Launch Pad profile\n" \
-             " surround arguments with spaces with single quotes'" \
-             "\n ADMIN/TEAM LEAD ONLY option: --member MEMBER_ID\n" \
-             " 'edit properties of another user's Launch Pad profile'\n\n" \
-             " view MEMBER_ID\n 'view information about a user'\n\n " \
-             "help\n 'outputs options for user commands'\n\n " \
-             "ADMIN ONLY\n\n delete MEMBER_ID\n" \
-             " 'permanently delete member's Launch Pad profile'"
```
cheukyin699: I think that this stuff (`__help` and `__command_name`) shouldn't be object attributes; they should be static constants. Put them back (preferably without the double-underscores)!

Issue #29, opened 2018-10-06
Add event handler method


rwblickhan:
Add a `core.py` (or some other reasonable name) to a `command/` directory (or some other reasonable name) that will expose an event handler method that will be sent the event JSON strings from the Slack client. For now it just has to accept these and deserialize them to into SlackEvent objects (as implemented in #28).

PR #68, opened 2018-10-30
Rmcreyes/#29 event handler

rmcreyes:
# Pull Request

## Description

Added `Core` class that encapsulates the `handle` method for each type of command.

## Testing

## Ticket(s)

Affects #29

### DIFF
--- a/command/core.py
+++ b/command/core.py
@@ -1,21 +1,23 @@
-from slackeventsapi import SlackEventAdapter
+"""Calls the appropriate handler depending on the event data."""
 from command.commands.user import UserCommand
-import os
 
-commands = {}
-commands["user"] = UserCommand()
-slack_signing_secret = os.environ["SLACK_SIGNING_SECRET"]
-slack_events_adapter = SlackEventAdapter(slack_signing_secret, "/slack/events")
-    
 
-@slack_events_adapter.on("app_mention")
-def handle_mention(event_data):
-    message = event_data["event"]["text"]
-    s = message.split(' ', 2)
-    if s[0] is not "@rocket":
-        pass
-    else:
-        command = s[1] + ' ' + s[2]
-        commands[s[1]].handle(command)
-    
-slack_events_adapter.start(port=3000)
+class Core:
+    """Encapsulate methods for handling events."""
+
+    def __init__(self):
+        """Initialize the dictionary of command handlers."""
+        self.__commands = {}
+        self.__commands["user"] = UserCommand()
+
+    def handle_app_mention(self, event_data):
+        """Handle the events associated with mentions of @rocket."""
+        message = (event_data["event"])["text"]
+        s = message.split(' ', 2)
+        if s[0] != "@rocket":
+            return False
+        else:
+            command_type = s[1]
+            command = command_type + ' ' + s[2]
+            self.__commands[command_type].handle(command)
+            return True
--- a/command/core.py
+++ b/command/core.py
@@ -1,10 +1,12 @@
 from slackeventsapi import SlackEventAdapter
+from command.commands.user import UserCommand
 import os
 
+commands = {}
+commands["user"] = UserCommand()
 slack_signing_secret = os.environ["SLACK_SIGNING_SECRET"]
 slack_events_adapter = SlackEventAdapter(slack_signing_secret, "/slack/events")
-
-commands = {}
+    
 
 @slack_events_adapter.on("app_mention")
 def handle_mention(event_data):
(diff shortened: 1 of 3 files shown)


command/core.py
```diff
@@ -0,0 +1,23 @@
+"""Calls the appropriate handler depending on the event data."""
+from command.commands.user import UserCommand
+
+
+class Core:
+    """Encapsulate methods for handling events."""
+
+    def __init__(self):
```
rwblickhan: I think eventually we'll want the command handler dictionary to be injected (or, if it's constructed locally, to have some way to add things to it from outside the class, such that we can add the commands from outside). I wouldn't bother changing it for this pull request, but keep it in mind for future ones.

command/core.py
```diff
@@ -0,0 +1,23 @@
...
+        """Initialize the dictionary of command handlers."""
+        self.__commands = {}
+        self.__commands["user"] = UserCommand()
+
+    def handle_app_mention(self, event_data):
+        """Handle the events associated with mentions of @rocket."""
+        message = (event_data["event"])["text"]
+        s = message.split(' ', 2)
+        if s[0] != "@rocket":
```
rwblickhan: We'll want to double check if this works once we get the actual Slack app running - I know the RTM API sends mentions with a Slack ID that looks something like `<@u19238>`, but I'm not sure that the Events API does as well. If it does, this line will break.

tests/command/core_test.py
```diff
@@ -0,0 +1,49 @@
...
+    event["event_id"] = "Ev08MFMKH6"
+    event["event_time"] = 1234567890
+    core = Core()
+    assert not core.handle_app_mention(event)
+
+
+@mock.patch('command.core.UserCommand')
+def test_handle_user_command(MockUserCommand):
+    """Test that UserCommand.handle is called appropriately."""
+    event = {}
```
rwblickhan: Same.

command/core.py
```diff
@@ -0,0 +1,23 @@
...
+            return False
+        else:
+            command_type = s[1]
+            command = command_type + ' ' + s[2]
+            self.__commands[command_type].handle(command)
```
cheukyin699: Should add a fail-safe (try-catch) for when an invalid command is used.

command/core.py
```diff
@@ -0,0 +1,23 @@
...
+    """Encapsulate methods for handling events."""
+
+    def __init__(self):
+        """Initialize the dictionary of command handlers."""
+        self.__commands = {}
+        self.__commands["user"] = UserCommand()
+
+    def handle_app_mention(self, event_data):
+        """Handle the events associated with mentions of @rocket."""
+        message = (event_data["event"])["text"]
```
cheukyin699: No need to use parentheses here.

tests/command/core_test.py
```diff
@@ -0,0 +1,49 @@
+"""Test the core event handler."""
+from unittest import mock
+from command.core import Core
+
+
+def test_handle_invalid_mention():
+    """Test the instance of handle_app_mention being called inappropriately."""
+    event = {}
```
rwblickhan: This can be made a bit cleaner using a dictionary literal, which would something look like this:

```python
event = {
    "token": "XXYYZZ",
    "team_id": "TXXXXXXXX",
    "api_app_id": "AXXXXXXX",
    "event": {
        "type": "app_mention",
        ...,
    },
    ...
}
```

command/core.py
rmcreyes: Ok I'll modify the function so it returns 0 on invalid mention, 1 on successful handle, and -1 on error. Is that ok?

command/core.py
rmcreyes: Ok I'll change that.

tests/command/core_test.py
rmcreyes: Ok cool, I'll write it like that

command/core.py
rmcreyes: I'm guessing this is something we'll have to look back at later then?

rwblickhan: Yeah, I don't think we can do it until we actually set it up to receive events from Slack (which I'll... try to get to this week 😂), but it might be helpful to put a TODO up there.

command/core.py
rwblickhan: Eventually we'll want it to log as well, though since we never got logging set up...

cheukyin699: @rwblickhan is it a good idea to throw exceptions here? and to let whoever to handle them? as in logging?

rwblickhan: Personally I think the best solution here would be to not return/throw anything, but rather to just log an error and maybe message the user that sent it (something like "Invalid command"), but seeing as how we don't have logging or Slack messages set up yet, that doesn't really work.

(I think we would want to avoid doing error handling in the Slack events handling module that will call this, so that that can remain as simple as possible.)

PR #74, opened 2018-11-03
Add dynamodb team API

cheukyin699:
# Pull Request

## Description

Add ability to store, retrieve, query, and delete teams from dynamodb.


tests/db/dynamodb_test.py
```diff
@@ -46,3 +46,35 @@ def test_query_user():
     assert user == strict_users[0]
 
     ddb.delete_user('abc_123')
+
+
+def test_store_retrieve_team():
```
rwblickhan: These should be marked as `db` once the other pull request is merged, correct?

cheukyin699: yes.

PR #75, opened 2018-11-03
Added user slack_id parameter in command parser and in core

jacqueschen1:
# Pull Request

## Description

Added user slack_id parameter in command parser and in core

## Testing

## Ticket(s)

Affects #


command/commands/user.py
```diff
@@ -28,7 +28,7 @@ class UserCommand:
     """Parser for view command."""
     parser_view = subparsers.add_parser("view")
     parser_view.set_defaults(which="view")
-    parser_view.add_argument("slack_id", type=str, action='store')
+    parser_view.add_argument("--slack_id", type=str, action='store')
```
rwblickhan: Since the Slack ID is required to use the `view` command, this probably shouldn't be a full flag but rather a required argument (though I'm not sure off the top of my head how to do that); that is, per the docs, the command should be `@rocket user view SLACK_ID`, but this will parse `@rocket user view --slack_id SLACK_ID`, which is slightly less convenient.

command/core.py
```diff
@@ -13,14 +13,15 @@ def __init__(self):
     def handle_app_mention(self, event_data):
         """Handle the events associated with mentions of @rocket."""
         message = event_data["event"]["text"]
+        user = event_data["event"]["user"]
```
rwblickhan: Once we have this running we'll want to double check what format `event_data["event"]["user"]` is (e.g. is it the actual `@rwblickhan`, or something like `<@U12345>`?).

rwblickhan:
The `.pyc`/`__pycache__` files shouldn't be committed. I think they're included in the .gitignore now so if you delete them and commit that it should work fine from now on (I can show you that if you'd like).

command/commands/user.py
jacqueschen1: I made it optional so that if the argument isn't included, then it'll just view the information of the user who called the command. The optional `--slack_id` is for if you want to view information about another user's info

PR #76, opened 2018-11-03
Pytest tests for dynamodb things is optional

### LINKED ISSUE #72: Research pytest and different types of tests (if possible)
Have `pytest` run database tests only when travis is running the tests (full tests). If the user doesn't have the database running (by pinging `localhost:8000`), we just give a warning, saying that the database isn't running, or have a whole different group of tests (unit tests vs functional tests).

cheukyin699:
# Pull Request

## Description

It still runs by default, but allows you the option of skipping/deselecting it if you run `pytest`.

## Ticket(s)

Closes #72

### DIFF
--- a/scripts/build_check.sh
+++ b/scripts/build_check.sh
@@ -3,4 +3,14 @@ set -euxo pipefail
 pipenv run pycodestyle .
 pipenv run pydocstyle .
 mdl .
-pipenv run pytest tests/
+
+# We use nmap to check if dynamodb is running locally
+# TODO: later should combine this with environmental variable checking (to see
+# if we can use a remote server instead).
+# XXX: Find a better way to check if dynamodb is running locally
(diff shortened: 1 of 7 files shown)


scripts/build_check.sh
```diff
@@ -3,4 +3,14 @@ set -euxo pipefail
...
-pipenv run pytest tests/
+
+# We use nmap to check if dynamodb is running locally
+# TODO: later should combine this with environmental variable checking (to see
+# if we can use a remote server instead).
+# XXX: Find a better way to check if dynamodb is running locally
+if nmap localhost | egrep "8000.*http-alt"; then
+    pipenv run pytest tests/
+else
+    printf "Warning: DynamoDB not detected. Running without the tests.\n"
```
rwblickhan: The warning could be clarified slightly; this makes it sound like it's not running tests at all.

PR #77, opened 2018-11-04
Setup continuous deployment

### LINKED ISSUE #24: Setup deployment/continuous delivery
Set up the very basics of deployment, probably with AWS EC2/DynamoDb, so that we can begin deploying after every pull request.

rwblickhan:
# Pull Request

## Description

Fix up the Dockerfile and create a Flask instance that for now serves a simple status page. This is being continuously deployed with [Inertia](https://github.com/ubclaunchpad/inertia).

## Testing

I've tested this by explicitly setting Inertia to deploy this branch and it seems to work, though it's now been reset to deploy the master branch (which should occur automatically once this is merged). Once it's deployed again I _believe_ the status page should show up at http://ec2-35-175-145-150.compute-1.amazonaws.com:5000.

## Ticket(s)

Closes #24

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.pyc
 __pycache__/
 .env
+DynamoDB/
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,10 +7,10 @@ WORKDIR /app
 COPY . /app
 
 # Install everything
-RUN apt-get update && apt-get install -y rubygems
+RUN apt-get update
 RUN pip install pipenv
 RUN pipenv install
-RUN gem install mdl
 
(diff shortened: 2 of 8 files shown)


rwblickhan:
What instructions do you think are missing? I was intending the scripts to serve as documentation - you just run `docker_build.sh` and then you can run `docker_run_local.sh`, though seeing as how it will ultimately require events from Slack that won't do much (we could definitely look into better ways to test - it's possible Inertia would allow us to have multiple branches deploying at the same time and we could set up a proxy to copy Slack events to each, or something like that - but that would be significantly more work).

PR #79, opened 2018-11-05
Add default handler for user

### LINKED ISSUE #63: Handle default command parsing for command classes
Currently, if a command such as `@rocket user` is handled without any subparsers, an error is returned. Add a default handler so that handler returns help text.

cheukyin699:
# Pull Request

## Description

Add default handler for user, and also add test for that one case.

## Ticket(s)

Closes #63

### DIFF
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -31,7 +31,7 @@ class UserCommand:
 
     def init_subparsers(self):
         """Initialize subparsers for user command."""
-        subparsers = self.parser.add_subparsers()
+        subparsers = self.parser.add_subparsers(dest="which")
 
         """Parser for view command."""
         parser_view = subparsers.add_parser("view")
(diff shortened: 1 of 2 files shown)


cheukyin699:
**WAIT WAIT WAIT I SWEAR IT WORKS ON MY MACHINE**

command/commands/user.py
```diff
@@ -69,8 +69,18 @@ def get_help(self):
...
         """Handle command by splitting string and giving to parser."""
         command_arg = shlex.split(command)
-        args = self.parser.parse_args(command_arg)
-        if args.which == "view":
+        args = None
+
+        try:
+            # Going for the nuclear option
+            args = self.parser.parse_args(command_arg)
+        except SystemExit:
```
rwblickhan: Is this safe? Could there be cases where `parser` legitimately wants to throw `SystemExit`?

cheukyin699: If I have my info correct, the only way for `parser.parse_args` to raise `SystemExit` is if any of the required parsed arguments are missing. I kept looking through post after post of people dissing the `argparse` library just because of its inflexibility, and this was one of the stack overflow "answers". Apparently, they made it so that there is no good way of handling the case where you don't provide correct arguments.

cheukyin699: The other (non-savoury) option would be to create a custom `ArgumentParser`, just for this thing, that overrides a specific function in the parent class.

PR #90, opened 2018-11-10
Add dynamodb store team validation

### LINKED ISSUE #86: Team storing isn't validated
It doesn't give any feedback on if the team is stored correctly or not. It should return true if the team is stored successfully, and false if it isn't.

cheukyin699:
# Pull Request

## Description

It used to be that dynamodb just checked the data in user storing, and not for team storing. **NOT ANYMORE**.

## Ticket(s)

Closes #86.


cheukyin699:
This has been a roller-coaster-ride of emotions....

PR #91, opened 2018-11-10
Update Team Command

### LINKED ISSUE #50: Create initial `team` command documentation
Create a document in `docs/` (similar to the one created for #25) that lists the various options/help strings/etc that we will implement for MVP for the `@rocket team` command.

In particular, we will want to match the functionality of the current `@rocket edit-team`, `@rocket remove-user`, `@rocket view-team`, `@rocket add-team`, `@rocket add-user`, `@rocket remove-team`, and `@rocket teams` commands (but hopefully with better/more consistent names!). These should probably all be implemented as subcommands to the core `@rocket team` command (similar to how user commands are being done).

### LINKED ISSUE #51: Stub out `team` command
Create a team command class in `command/commands/`, stub out its main methods (e.g. `handle()`), and implement `help()` and `name()` (double check with the implementation of the user command as implemented in #31).

### LINKED ISSUE #52: Implement `team` command parsing
In the `handle()` command created in #51, implement argument parsing with [`argparse`](https://docs.python.org/3/library/argparse.html). This should be based on the documentation created in #50.

vyzw:
# Pull Request

## Description

- Add help text to Team Command and test

## Testing

## Ticket(s)

Closes #50 
Closes #51 
Closes #52 


command/commands/team.py
```diff
@@ -7,7 +7,31 @@ class TeamCommand:
...
+    * `edit` GITHUB_TEAM_NAME
+        * [`--name` DISPLAY_NAME]
+        * [`--platform` PLATFORM]
+            * edit properties of specified team
+    * `add` GITHUB_TEAM_NAME @Slack User
+        * add the specified user to the team
+    * `remove` GITHUB_TEAM_NAME @Slack User
+        * remove the specified user from the team
+    * `delete` GITHUB_TEAM_NAME
+        * permanently delete the specified team"""
```
cheukyin699: Please un-indent the string; the indents would carry over, making the string that gets passed to slack a bit padded (and even more importantly, ugly).

cheukyin699: Would a better way be to read the `docs/TeamCommand.md` file and just dump it into the help string? A bit of a dilemma here, that you would have to choose between modifying both strings when the docs change, or reading and copying the text of a docs file.

tests/command/commands/team_test.py
```diff
@@ -2,10 +2,34 @@
...
+    * `edit` GITHUB_TEAM_NAME
+        * [`--name` DISPLAY_NAME]
+        * [`--platform` PLATFORM]
+            * edit properties of specified team
+    * `add` GITHUB_TEAM_NAME @Slack User
+        * add the specified user to the team
+    * `remove` GITHUB_TEAM_NAME @Slack User
+        * remove the specified user from the team
+    * `delete` GITHUB_TEAM_NAME
+        * permanently delete the specified team"""
```
cheukyin699: `help_text` should be the same as `TeamCommand.help`. Should be being the keyword. So maybe use that instead of having a copy of the `help_text`.

command/commands/team.py
vyzw: I was trying to grab the text from the file, but I was confused at what the path should be. 
`help = open("docs/TeamCommand.md", "r").read()` works when you run pytest in the rocket2.0 folder, but not when you run pytest in the test folder, for example.

docs/TeamCommands.md
```diff
@@ -28,10 +28,10 @@ The following can only be used by a team lead in the team or an admin:
   * [`--platform` PLATFORM]
     * edit properties of specified team
 
-* `add` GITHUB_TEAM_NAME SLACK_ID
+* `add` GITHUB_TEAM_NAME @Slack User
```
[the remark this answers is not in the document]
joy8zhang: Should we consider a way to add multiple slack users at once ie. add all the members of the teams in one line v.s. calling 1 line per member.

tests/command/commands/team_test.py
```diff
@@ -44,8 +68,12 @@ def test_handle_delete():
 def test_handle_create():
     """Test team command create parser."""
     testcommand = TeamCommand()
-    assert testcommand.handle("team create b-s") == "id b-s"
-    assert testcommand.handle("team create b-s --name B") == "team B, id b-s"
+    inputstring = "team create b-s --name 'B S'"
+    outputstring = "new team: b-s, name: B S, "
+    assert testcommand.handle(inputstring) == outputstring
+    inputstring += " --platform web"
+    outputstring += "platform: web, "
```
joy8zhang: Is a completed line of outputstring intended to end with a `,`?

command/commands/team.py
cheukyin699: Good point

tests/command/commands/team_test.py
vyzw: That's how the method is currently written. I'm okay with this for now, because we're only testing parsing.

PR #94, opened 2018-11-11
Read SLACK_SIGNING_SECRET from env var

### LINKED ISSUE #92: Server crashes with NameError on launch
`SLACK_SIGNING_SECRET` is undefined, it should be read in from an environment variable.
```
File "/Users/rwblickhan/Developer/rocket2.0/server/server.py", line 16, in <module>
    slack_events_adapter = SlackEventAdapter(SLACK_SIGNING_SECRET,
NameError: name 'SLACK_SIGNING_SECRET' is not defined
[2018-11-10 15:44:14 -0800] [715] [INFO] Worker exiting (pid: 715)
```

rwblickhan:
# Pull Request

## Description

Read `SLACK_SIGNING_SECRET` from env var to avoid server crashes.

## Ticket(s)

Closes #92

### DIFF
--- a/server/server.py
+++ b/server/server.py
@@ -2,6 +2,7 @@
 from flask import Flask
 from slackeventsapi import SlackEventAdapter
 from command.core import Core
+import os
 
 app = Flask(__name__)
 core = Core()
@@ -13,6 +14,7 @@ def check():
     return "🚀"
 
 
+SLACK_SIGNING_SECRET = os.environ["SLACK_SIGNING_SECRET"]
 slack_events_adapter = SlackEventAdapter(SLACK_SIGNING_SECRET,
                                          "/slack/events", app)
 


server/server.py
```diff
@@ -13,6 +14,7 @@ def check():
     return "🚀"
 
 
+SLACK_SIGNING_SECRET = os.environ["SLACK_SIGNING_SECRET"]
```
cheukyin699: There is still a chance that everything will crash (though smaller likelihood). I think you should use `os.environ.get("SLACK_SIGNING_SECRET", "")` so that we don't have a key error and can specify a default value.

rwblickhan: A good point, but I wonder if there's a value in instead simply logging+crashing when `SLACK_SIGNING_SECRET` isn't found (which would, of course, still require that I catch the `KeyError`). After all, it should _always_ be available on launch, and if we specify a default we'll get all kinds of errors later on that might be more difficult to debug.

PR #95, opened 2018-11-12
added basic logging

### LINKED ISSUE #42: Setup logging
After deployment we'll want to set up some logging. In particular we will want logging levels as well as a general facade structure (so that for instance we can use a local log file or AWS CloudWatch Logs in the future). These should be provided by the [`logging`](https://docs.python.org/3.7/library/logging.html) standard library module - see also the [logging tutorial](https://docs.python.org/3.7/howto/logging.html#logging-basic-tutorial) and [tips on Hitchhiker's Guide](https://docs.python-guide.org/writing/logging/).

rmcreyes:
# Pull Request

## Description

Configured the logging module in `server.py`. Any logging that needs to be done in other modules just needs to import the `logging` module and call the appropriate logging level function. Also removed `SLACK_SIGNING_SECRET` while it is left undefined, but can still change it back before the final merge if need be.

## Testing

Tested just by running the server with `pipenv run launch`

## Ticket(s)

Closes #42 

### DIFF
--- a/server/server.py
+++ b/server/server.py
@@ -2,18 +2,22 @@
 from flask import Flask
 from slackeventsapi import SlackEventAdapter
 from command.core import Core
+import logging
 
 app = Flask(__name__)
 core = Core()
+logging.basicConfig(format='%(asctime)s - %(levelname)s : %(message)s',
+                    level=logging.INFO)
 
 
 @app.route('/')
 def check():
     """Display a Rocket status image."""
+    logging.info('ROCKET2 IS RUNNING')
     return "🚀"
 
 
-slack_events_adapter = SlackEventAdapter(SLACK_SIGNING_SECRET,
+slack_events_adapter = SlackEventAdapter(None,
                                          "/slack/events", app)
 
 
--- a/server/server.py
+++ b/server/server.py
@@ -18,7 +18,6 @@ def check():
     return "🚀"
 
 
-
 SLACK_SIGNING_SECRET = os.environ["SLACK_SIGNING_SECRET"]
 slack_events_adapter = SlackEventAdapter(SLACK_SIGNING_SECRET,
                                          "/slack/events", app)
--- a/server/server.py
+++ b/server/server.py
@@ -7,14 +7,15 @@ import logging
 
 app = Flask(__name__)
 core = Core()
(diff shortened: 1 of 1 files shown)


cheukyin699:
Think there's a separate PR for removing `SLACK_SIGNING_SECRET`, which is in the process of review. Don't remove it here.

server/server.py
```diff
@@ -2,18 +2,22 @@
...
 
 
 @app.route('/')
 def check():
     """Display a Rocket status image."""
+    logging.info('ROCKET2 IS RUNNING')
     return "🚀"
 
 
-slack_events_adapter = SlackEventAdapter(SLACK_SIGNING_SECRET,
```
rwblickhan: Like @cheukyin699 said, PR #94 should fix this, so this should be left in place for this PR (or we'll get all kinds of nasty merge conflicts :joy:).

server/server.py
```diff
@@ -2,18 +2,22 @@
...
 app = Flask(__name__)
 core = Core()
+logging.basicConfig(format='%(asctime)s - %(levelname)s : %(message)s',
+                    level=logging.INFO)
 
 
 @app.route('/')
 def check():
     """Display a Rocket status image."""
+    logging.info('ROCKET2 IS RUNNING')
```
rwblickhan: This function gets called every time the server serves a request to `/`, so I think this log message could be a bit more informative (e.g. `"Served 'check()'"` or `"Served route '/'"`), and the general "Rocket 2 startup log" should be placed outside (e.g. right below where logging is configured).

Also, just a nit, but I would prefer not to have all-caps logs - if all the logs are all-caps it'll be much harder to read in a console when debugging.

server/server.py
```diff
@@ -2,18 +2,22 @@
 from flask import Flask
 from slackeventsapi import SlackEventAdapter
 from command.core import Core
+import logging
 
 app = Flask(__name__)
 core = Core()
+logging.basicConfig(format='%(asctime)s - %(levelname)s : %(message)s',
```
rwblickhan: Would you add `%(module)s` and maybe `%(funcName)s` to the config? That would add the module and function doing the calling, which would be more convenient for debugging.

Issue #64, opened 2018-10-29
Start implementation of `user` Commands with database


jacqueschen1:
Create basic implementation between user commands and database for user edit, view, and delete commands

rwblickhan:
Closing - this should be closed by #100.

PR #100, opened 2018-11-14
#64 user command database implementation

jacqueschen1:
# Pull Request

## Description
User command view and delete commands implementation and tests
Changed core and usercommand constructors to include bot and database facade parameters

## Testing

## Ticket(s)

Affects #64 

### DIFF
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -3,6 +3,7 @@ import argparse
 import shlex
 from db.facade import DBFacade
 from db.dynamodb import DynamoDB
+from model.permissions import Permissions
 
 
 class UserCommand:
@@ -27,11 +28,13 @@ class UserCommand:
     permission_error = "You do not have the sufficient " \
                        "permission level for this command!"
 
-    def __init__(self):
-        """Initialize user command parser."""
+    def __init__(self, DBfacade, bot):
+        """Initialize user command."""
         self.parser = argparse.ArgumentParser(prog="user")
         self.parser.add_argument("user")
         self.init_subparsers()
+        self.facade = DBfacade
+        self.bot = bot
 
     def init_subparsers(self):
         """Initialize subparsers for user command."""
@@ -70,7 +73,7 @@ class UserCommand:
         """Return command options for user events."""
         return self.help
 
-    def handle(self, command, user_id):
+    def handle(self, command, user_id, channel):
         """Handle command by splitting into substrings and giving to parser."""
         command_arg = shlex.split(command)
         args = None
@@ -98,7 +101,7 @@ class UserCommand:
         elif args.which == "delete":
             # stub
             # return "deleting " + args.slack_id
-            return self.delete_helper(user_id, args.slack_id)
+            return self.delete_helper(user_id, args.slack_id, channel)
 
         elif args.which == "edit":
             # stub
@@ -119,7 +122,7 @@ class UserCommand:
                 msg += "bio: {}".format(args.bio)
             return msg
 
-    def delete_helper(self, user_id, slack_id):
+    def delete_helper(self, user_id, slack_id, channel):
         """
         Delete user from database.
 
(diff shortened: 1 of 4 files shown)


command/commands/user.py
```diff
@@ -22,12 +25,17 @@ class UserCommand:
            "help\n 'outputs options for user commands'\n\n " \
            "ADMIN ONLY\n\n delete MEMBER_ID\n" \
            " 'permanently delete member's Launch Pad profile'"
+    permission_error = "You do not have the sufficient " \
+                       "permission level for this command!"
+    lookup_error = "User not found!"
 
-    def __init__(self):
-        """Initialize user command parser."""
+    def __init__(self, DBfacade, bot):
```
rwblickhan: Nit: snake case would have this named as `db_facade`.

command/commands/user.py
```diff
@@ -113,3 +119,50 @@ def handle(self, command, user_id):
...
+        Delete user with slack_id from database if user with user_id has
+        admin permission level.
+
+        :param user_id: Slack ID of user who is calling the command
+        :param slack_id: Slack ID of user who is being deleted
+        :param channel: ID of Slack channel that called command
+        :return: returns permission error message if not admin,
+                 returns deletion message if user is deleted.
+        """
+        message = "Deleted user with Slack ID: " + slack_id
```
rwblickhan: I’m not sure it’s worth having `message` as a local variable, seeing as how it’s only used in one place (though I’d be happy to hear arguments for/against).

command/core.py
```diff
@@ -5,23 +5,26 @@
 class Core:
     """Encapsulate methods for handling events."""
 
-    def __init__(self):
+    def __init__(self, DBfacade, bot):
```
rwblickhan: Nit: snake case.

command/core.py
```diff
@@ -5,23 +5,26 @@
...
         """Initialize the dictionary of command handlers."""
         self.__commands = {}
-        self.__commands["user"] = UserCommand()
+        self._facade = DBfacade
```
rwblickhan: Nit: looks like `facade` only has one preceding underscore.

command/core.py
```diff
@@ -5,23 +5,26 @@
...
+        self.__bot = bot
+        self.__commands["user"] = UserCommand(self._facade, self.__bot)
```
rwblickhan: This is fine for now but I think we’ll want a more principled way to do this in the future (e.g. for once we add the Github integration - that’ll need to be dependency injected too); I have some thoughts in that direction we can talk about this weekend.

tests/command/commands/user_test.py
```diff
@@ -1,66 +1,174 @@
...
+from unittest import mock
+from model.user import User
+from db.facade import DBFacade
+from model.permissions import Permissions
+from bot.bot import Bot
 
 
 def test_get_command_name():
     """Test user command get_name method."""
-    testcommand = UserCommand()
```
rwblickhan: A lot of this setup is shared between unit tests - this can be cleaned up a bit by using test cases/test fixtures, which are provided by both [pytest](
http://doc.pytest.org/en/latest/fixture.htm) and via [`unittest`](
https://docs.python.org/3/library/unittest.htm), which can be run by Pytest. I believe the [`bot` tests](
https://github.com/ubclaunchpad/rocket2.0/blob/master/tests/bot/bot_test.py) are already in the `unittest` style.

command/commands/user.py
```diff
@@ -75,25 +83,23 @@ def handle(self, command, user_id):
...
+            return self.bot.send_to_channel(self.help, channel)
 
         elif args.which == "view":
             # stub
             if args.slack_id is not None:
-                return args.slack_id
+                self.view_helper(user_id, args.slack_id, channel)
             else:
-                return user_id
+                self.view_helper(user_id, None, channel)
```
rwblickhan: These two calls to `view_helper` can be squashed together (the it statement doesn’t really do anything).

command/commands/user.py
cheukyin699: Since these are constants, they should be in the static scope. Something like `self.delete_text` or `self.delete_status`.

rwblickhan: We may eventually want to read them from a config file as well - storing static strings in config is quite common in mobile development. That should definitely be a future PR though.

command/commands/user.py
```diff
@@ -1,6 +1,9 @@
 """Command parsing for user events."""
 import argparse
 import shlex
+from db.facade import DBFacade
+from db.dynamodb import DynamoDB
```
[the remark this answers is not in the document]
jacqueschen1: my b

PR #105, opened 2018-11-17
fixed core init in server

jacqueschen1:
# Pull Request

## Description
fixed core init in server
## Testing

## Ticket(s)

Affects #


rwblickhan:
Do not merge this yet, we may want to handle this in another way - also we need to deal with connecting to DynamoDb for tests vs. in prod.

cheukyin699:
Blocked until #106 is done, if we are going with that route.

rwblickhan:
Closing this PR without merging - this has been handled by #122 and #108.

PR #108, opened 2018-11-17
Testing enviornment variable for dynamodb

### LINKED ISSUE #106: Differentiate production and testing environments
We have an issue with DynamoDB: it dies when we use a non-local DynamoDB instance (the one on Amazon). We need it to not do that. We want to differentiate between production and testing/development environments, preferably using environmental variables. There should be a change in the `db/dynamodb.py` file `__init__` function:

```python
import os
def __init__(self):
    # Default to testing if no env variable
    TESTING = bool(os.environ.get('TESTING', True))
    if TESTING:
        self.ddb = boto3.resource('dynamodb', endpoint_url='localhost:8000')
    else:
        self.ddb = boto3.resource('dynamodb')
```

Please specify in some form of documentation that we will be using `TESTING` as an environmental variable, and how we use it.

Note that we want pass environmental variables into the application. Please document what kind of variables are needed. [Here is the link][boto3] to get you started. We'd also need to pass the non-credentials such as region.

Please use the above code as a template and change as necessary.

[boto3]: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#environment-variables

joy8zhang:
# Pull Request

## Description
Added testing environment variable so local tests won't fail when we use non-local dynamodb instance.

## Testing

For testing, we change s.t. we use `TESTING=True pytest` instead of just `pytest`.

## Ticket(s)

Closes #106

### DIFF
--- a/db/dynamodb.py
+++ b/db/dynamodb.py
@@ -39,7 +39,7 @@ class DynamoDB:
         AWS_SESSION_TOKEN: The session key for your AWS account.
         This is only needed when you are using temporary credentials.
         """
-        testing = bool(os.environ.get("TESTING"), True)
+        testing = bool(os.environ.get("TESTING", True))
 
         if testing:
             self.ddb = boto3.resource("dynamodb",
--- a/db/dynamodb.py
+++ b/db/dynamodb.py
@@ -1,5 +1,6 @@
 """DynamoDB."""
 import boto3
+import logging
 import os
 from boto3.dynamodb.conditions import Attr
 from model.user import User
@@ -40,13 +41,18 @@ class DynamoDB:
         This is only needed when you are using temporary credentials.
         """
         testing = bool(os.environ.get("TESTING", True))
+        region_name = os.environ.get("REGION", '')
+        if region_name == '':
+            region_name = 'us-east-1'
 
         if testing:
+            logging.info('Local Dynamodb is running')
             self.ddb = boto3.resource("dynamodb",
                                       region_name="",
                                       endpoint_url="http://localhost:8000")
         else:
-            self.ddb = boto3.resource('dynamodb', region_name='us-west-1')
+            logging.info('Server Dynamodb is running')
+            self.ddb = boto3.resource('dynamodb', region_name=region_name)
 
         if not self.check_valid_table('users'):
             self.__create_user_tables()
--- a/db/dynamodb.py
+++ b/db/dynamodb.py
@@ -19,10 +19,12 @@ class DynamoDB:
     def __init__(self):
         """Initialize facade using DynamoDB settings.
 
-        To avoid local tests failure when the dyanmodb server is used, a testing
-        environment variable is set.
-        When testing environmental variable is true, the local dynamodb is run.
-        When testing environmental variable is true, the server dynamodb is run.
+        To avoid local tests failure when the dyanmodb server is used,
+        a testing environment variable is set.
+        When testing environmental variable is true,
+        the local dynamodb is run.
+        When testing environmental variable is true,
+        the server dynamodb is run.
 
         boto3.resource() takes in a service_name, region_name, and endpoint_url
         (only for local dynamodb).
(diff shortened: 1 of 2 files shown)


db/dynamodb.py
```diff
@@ -16,10 +17,36 @@ class DynamoDB:
...
+        This is only needed when you are using temporary credentials.
+        """
+        testing = bool(os.environ.get("TESTING", True))
+
+        if testing:
+            self.ddb = boto3.resource("dynamodb",
+                                      region_name="",
+                                      endpoint_url="http://localhost:8000")
+        else:
+            self.ddb = boto3.resource('dynamodb', region_name='us-west-1')
```
rwblickhan: `region_name` should probably also be read from an environment variable (or a config file, but we don't have any configs yet...). In particular, we're running on `us-east-1` at the moment.

db/dynamodb.py
```diff
@@ -16,10 +17,36 @@ class DynamoDB:
...
+        A list of different regions can be obtained online.
+        endpoint_url: The complete URL to use for the constructed client.
+
+        Boto3 server require environmental variables for credentials:
+        AWS_ACCESS_KEY_ID: The access key for your AWS account.
+        AWS_SECRET_ACCESS_KEY: The secret key for the AWS account
+        AWS_SESSION_TOKEN: The session key for your AWS account.
+        This is only needed when you are using temporary credentials.
+        """
+        testing = bool(os.environ.get("TESTING", True))
```
rwblickhan: How does this work? I'm having trouble understanding it.

db/dynamodb.py
```diff
@@ -16,10 +17,36 @@ class DynamoDB:
...
+
+        if testing:
+            self.ddb = boto3.resource("dynamodb",
```
rwblickhan: Oh, one thing I forgot - it's probably a good idea to log here (and in the else case). I would recommend using the `INFO` logging level.

db/dynamodb.py
cheukyin699: If we find the environmental variable "TESTING", we use it, otherwise, we default to `TESTING = True`.

rwblickhan: Doh! Something else I forgot - if the `TESTING` environment variable has to be there to run, it also needs to be added to the [docker-compose.yml](https://github.com/ubclaunchpad/rocket2/blob/master/docker-compose.yml) - you can follow the format there or (since we are deploying to production) you could set it directly:
```
environment:
            VIRTUAL_HOST: "${VIRTUAL_HOST}"
            ...
            TESTING: False
```

db/dynamodb.py
```diff
@@ -16,10 +18,41 @@ class DynamoDB:
...
+
+        Boto3 server require environmental variables for credentials:
+        AWS_ACCESS_KEY_ID: The access key for your AWS account.
+        AWS_SECRET_ACCESS_KEY: The secret key for the AWS account
+        AWS_SESSION_TOKEN: The session key for your AWS account.
+        This is only needed when you are using temporary credentials.
+        """
+        testing = bool(os.environ.get("TESTING", True))
+        region_name = os.environ.get("REGION", '')
+        if region_name == '':
```
rwblickhan: This can be squashed into the call to `os.environ.get()` - the second parameter is the default value to use.

db/dynamodb.py
```diff
@@ -16,10 +18,41 @@ class DynamoDB:
...
+        endpoint_url: The complete URL to use for the constructed client.
+
+        Boto3 server require environmental variables for credentials:
+        AWS_ACCESS_KEY_ID: The access key for your AWS account.
+        AWS_SECRET_ACCESS_KEY: The secret key for the AWS account
+        AWS_SESSION_TOKEN: The session key for your AWS account.
+        This is only needed when you are using temporary credentials.
+        """
+        testing = bool(os.environ.get("TESTING", True))
+        region_name = os.environ.get("REGION", '')
```
rwblickhan: Make sure to add `REGION` to the docker-compose.yml as well. Probably it should be read from the host environment (so something like `REGION: "${REGION}"`).

cheukyin699: Should also have the region default to something, since not doing so would probably error.

docker-compose.yml
```diff
@@ -27,6 +27,7 @@ services:
             VIRTUAL_HOST: "${VIRTUAL_HOST}"
             LETSENCRYPT_HOST: "${LETSENCRYPT_HOST}"
             LETSENCRYPT_EMAIL: "${LETSENCRYPT_EMAIL}"
-            SLACK_SIGNING_SECRET: "${SLACK_SIGNING_SECRET}"
+            SLACK_SIGNING_SECRET: "${SLACK_SIGNING_SECRET}
```
rwblickhan: Missing `"`! I'll go ahead and push that as an update so this can get merged right away.

PR #119, opened 2018-11-24
Rmcreyes/#73 user create handler

### LINKED ISSUE #73: Add slack_event handler for when user joins Slack and add User create handler
Add functionality such that a user is created and store to database when they first join Slack

rmcreyes:
# Pull Request

## Description

Added a method in `Core` that is meant to be called during the `team_join` event. Has not been called yet in `server.py`.

## Testing

N/A

## Ticket(s)

Closes #73 

### DIFF
--- a/command/core.py
+++ b/command/core.py
@@ -35,11 +35,10 @@ class Core:
         """Handle the event of a new user joining the workspace."""
         new_user_id = event_data["event"]["user"]["id"]
         new_user = User(new_user_id)
-        stored = self.__facade.store_user(new_user)
-        if stored:
-            welcome = 'Welcome to Lauchpad!'
-            try:
-                self.__bot.send_dm(welcome, new_user_id)
-                return True
-            except SlackAPIError:
-                return False
+        self.__facade.store_user(new_user)
+        welcome = 'Welcome to Lauchpad!'
+        try:
+            self.__bot.send_dm(welcome, new_user_id)
+            return True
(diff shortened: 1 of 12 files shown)


command/core.py
```diff
@@ -28,3 +30,15 @@ def handle_app_mention(self, event_data):
                 return 1
             except KeyError:
                 return -1
+
+    def handle_member_join(self, event_data):
```
rwblickhan: In the future we will want to refactor this into an `Event` class a la the `Command` classes. For now we can go ahead and merge this - this seems like most of the business logic we'll want.

command/core.py
```diff
@@ -28,3 +30,15 @@ def handle_app_mention(self, event_data):
...
+        """Handle the event of a new user joining the workspace."""
+        new_user_id = event_data["event"]["user"]["id"]
+        new_user = User(new_user_id)
+        self.__facade.store_user(new_user)
+        welcome = 'Welcome to Lauchpad!'
+        try:
+            self.__bot.send_dm(welcome, new_user_id)
+            return True
```
rwblickhan: A small thing - for the other handlers it looks like you used 1 and -1, but here it's using `True` and `False`. We should probably be consistent, though I also know we have #68 to get rid of these anyway.

rwblickhan:
@rmcreyes Can you create a ticket to add a Slack event handler that calls this?

PR #124, opened 2018-11-24
Make servers work again

### LINKED ISSUE #122: Core Factory for Core initialization
Right now the server isn't working correctly. We need to initialize `Core` class. So let's make a class/function that does such a thing!

cheukyin699:
# Pull Request

## Description

Add a factory to help initialize `Core` for the server (because imports are the devil in `server/server.py`). Also adds documentation for it and `Bot` (because I forgot it last time I did docs for some odd and inexplicable reason).

## Ticket(s)

Closes #122 .

### DIFF
--- a/factory/__init__.py
+++ b/factory/__init__.py
@@ -8,7 +8,7 @@ import os
 
 def make_core():
     """
-    Initializes and returns a :class:`command.core.Core` object.
+    Initialize and returns a :class:`command.core.Core` object.
 
     Requires that environmental variable ``SLACK_API_TOKEN`` to be set. Please
     set it to be the slack API token.
--- a/factory/__init__.py
+++ b/factory/__init__.py
@@ -1,0 +1,21 @@
+"""All necessary class initializations."""
+from db.facade import DBFacade
+from db.dynamodb import DynamoDB
+from command.core import Core
+from slackclient import SlackClient
+import os
+
+
+def make_core():
+    """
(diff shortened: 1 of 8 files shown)


factory/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+def make_core():
+    """
+    Initialize and returns a :class:`command.core.Core` object.
+
+    Requires that environmental variable ``SLACK_API_TOKEN`` to be set. Please
+    set it to be the slack API token.
+
+    :return: a new ``Core`` object, freshly initialized
+    """
+    slack_token = os.environ['SLACK_API_TOKEN']
```
rwblickhan: We may want to consider starting to put some of these into a config file, instead of stuffing them all in env vars - that wouldn't have to be in this PR (unless you want to do it...) but I think this would be the place to read everything from the config.

factory/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+from command.core import Core
+from slackclient import SlackClient
+import os
+
+
+def make_core():
+    """
+    Initialize and returns a :class:`command.core.Core` object.
+
+    Requires that environmental variable ``SLACK_API_TOKEN`` to be set. Please
```
rwblickhan: `DynamoDb` also requires an environment variable to be read, right? That should be noted here, and we might want to consider making that explicit - as in, pass in the value of that environment variable into the DynamoDb constructor. (Thoughts, anyone?)

PR #133, opened 2018-12-13
Add ngrok development tutorial

### LINKED ISSUE #130: Write tutorial on how to use ngrok with slack
Just do it please. So that we'd all live a little better.

### LINKED ISSUE #131: @rocket user view displaying json
It should not be displaying json. It should be displaying actual readable text. Please fix.

cheukyin699:
# Pull Request

## Description

Add ngrok tutorial. Fix bug where bot responds in json and not pretty pretty markdown. Fix the bug where it checks if `@rocket` is the first item in a command, when it could be literally any username. Now it has a slight bug where if you mention it and have any other username as the first element, it also works. It's a false positive, but completely work in progress I guess.

## Testing

If you would really like to, you can follow along with the tutorial (figuratively or literally) and see how far you'd get!

[Here is the link][lnk] for the lazies among us.

## Ticket(s)

Closes #131, closes #130.

[lnk]: https://rocket20.readthedocs.io/en/cheukyin699--130-ngrok-stuffs/docs/ServingTheServer.html

### DIFF
--- a/command/core.py
+++ b/command/core.py
@@ -21,14 +21,14 @@ class Core:
         user = event_data["event"]["user"]
         channel = event_data["event"]["channel"]
         s = message.split(' ', 2)
-        if s[0] != "@Rocket":
+        if s[0] != "@rocket":
             logging.error("app mention event triggered incorrectly")
         else:
             command_type = s[1]
             command = command_type + ' ' + s[2]
             try:
                 self.__commands[command_type].handle(command, user, channel)
-                logging.info(("@Rocket mention - "
+                logging.info(("@rocket mention - "
                               "successfully handled request: ") + message)
             except KeyError:
                 error_dm = "Please enter a valid command."
--- a/command/core.py
+++ b/command/core.py
@@ -17,7 +17,7 @@ class Core:
 
     def msg_split(self, msg):
         """
-        Splits a message into 3 components.
+        Split a message into 3 components.
 
         The 3 components, in order, are:
 
@@ -39,7 +39,7 @@ class Core:
 
         :return: A list of 3 strings
         """
-        return message.split(' ', 2)
+        return msg.split(' ', 2)
 
     def handle_app_mention(self, event_data):
         """Handle the events associated with mentions of @rocket."""
--- a/command/core.py
+++ b/command/core.py
@@ -20,6 +20,10 @@ class Core:
         message = event_data["event"]["text"]
         user = event_data["event"]["user"]
         channel = event_data["event"]["channel"]
+
+        if not message.startswith('<@U') and not message.startswith('@'):
+            return 0
+
         s = message.split(' ', 2)
         if s[0] != "@rocket":
             logging.error("app mention event triggered incorrectly")
--- a/command/core.py
+++ b/command/core.py
@@ -15,16 +15,39 @@ class Core:
         self.__bot = bot
         self.__commands["user"] = UserCommand(self.__facade, self.__bot)
 
+    def msg_split(self, msg):
+        """
+        Splits a message into 3 components.
+
+        The 3 components, in order, are:
+
+        - Mentioned bot slack ID
+        - Module of command
+        - Submodule and the rest of command
+
+        So the string ``<@ABC123> user view`` will be parsed into
+
+        .. code-block:: python
+
+            ['<@ABC123>', 'user', 'view']
+
+        And the string ``<@ABC123> user view @me`` will be parsed into
+
+        .. code-block:: python
+
+            ['<@ABC123>', 'user', 'view @me']
+
+        :return: A list of 3 strings
+        """
+        return message.split(' ', 2)
+
     def handle_app_mention(self, event_data):
         """Handle the events associated with mentions of @rocket."""
         message = event_data["event"]["text"]
         user = event_data["event"]["user"]
         channel = event_data["event"]["channel"]
 
-        if not message.startswith('<@U') and not message.startswith('@'):
-            return 0
-
-        s = message.split(' ', 2)
+        s = self.msg_split(message)
         if s[0] != "@rocket":
             logging.error("app mention event triggered incorrectly")
         else:
@@ -37,7 +60,7 @@ class Core:
             except KeyError:
                 error_dm = "Please enter a valid command."
                 self.__bot.send_dm(error_dm, user)
-                logging.info("@Rocket mention - invalid request: " + message)
+                logging.info("@rocket mention - invalid request: " + message)
 
     def handle_team_join(self, event_data):
         """Handle the event of a new user joining the workspace."""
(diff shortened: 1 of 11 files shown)


model/user.py
```diff
@@ -17,6 +17,25 @@ def __init__(self, slack_id):
         self.__image_url = ""
         self.__permissions_level = Permissions.member
 
+    def get_md_table(self):
+        """Return markdown table for user."""
+        form = """```
```
bobheadxi: just a thought - instead of using markdown tables for everything, Slack offers a very flexible [attachments API](https://api.slack.com/docs/message-attachments)

cheukyin699: That looks really nice! Will check it out.

cheukyin699: Like this instead?

image

bobheadxi: Hey that looks dope I love it!

command/core.py
```diff
@@ -19,17 +20,19 @@ def handle_app_mention(self, event_data):
         message = event_data["event"]["text"]
         user = event_data["event"]["user"]
         channel = event_data["event"]["channel"]
-        s = message.split(' ', 2)
-        if s[0] != "@rocket":
+
+        if not message.startswith('<@U') and not message.startswith('@'):
```
rwblickhan: The "right way" to solve this is probably to lookup the encoded ID for Rocket on startup (which should be added to our Slack interface) and then checking for that. Probably not necessary to do that for this PR, but good to keep in mind.

command/core.py
```diff
@@ -19,17 +20,19 @@ def handle_app_mention(self, event_data):
...
-        else:
-            command_type = s[1]
-            command = command_type + ' ' + s[2]
-            try:
-                self.__commands[command_type].handle(command, user, channel)
-                return 1
-            except KeyError:
-                return -1
+
+        s = message.split(' ', 2)
```
rwblickhan: Could some of this be factored out? I assume we'll have essentially the same logic for other event types (e.g. DMs vs channel messages).

command/core.py
```diff
@@ -19,17 +20,19 @@ def handle_app_mention(self, event_data):
...
+        command_type = s[1]
+        logging.info('{}:{}'.format(user, message))
+        command = command_type + ' ' + s[2]
+        try:
+            self.__commands[command_type].handle(command, user, channel)
+            return 1
+        except KeyError:
+            return -1
```
rwblickhan: I think we just want this to log and maybe send a message, not return, right?

factory/__init__.py
```diff
@@ -17,5 +18,5 @@ def make_core():
     """
     slack_token = os.environ['SLACK_API_TOKEN']
     facade = DBFacade(DynamoDB())
-    slack_client = SlackClient(slack_token)
-    return Core(facade, slack_client)
+    bot = Bot(SlackClient(slack_token))
```
rwblickhan: I wonder if we should rename the `Bot` class - it's really a general interface for the Slack Web API, not anything specifically bot-related.

docs/ServingTheServer.md
```diff
@@ -0,0 +1,138 @@
...
+After that, you must create a new slack app. Follow the link
+[here][make-slack-app] to create a new slack app. Make sure to have it installed
+onto your slack workspace of choice.
+
+Now, you need to create a bot user. [Create it here][make-bot].
+
+## 2: Download and Start Up DynamoDB
+
+Since we are developing locally, we must use the local version of DynamoDB.
+Download the `jar` from [here][dynamodb-download]. Follow the instructions from
```
rwblickhan: `dynamodb-download` link doesn't exist (looks like you named it `dynamodb` below).

docs/ServingTheServer.md
```diff
@@ -0,0 +1,138 @@
...
+
+## 3: Download and Install (and run) ngrok
+
+Slack does not run locally. To force/trick it to run locally (method recommended
+by slack), you must forward your 5000 port to a public domain name. Luckily for
+us, `ngrok` can do such things. Sign up and download and run `ngrok`
+[here][download-ngrok].
+
+Our app runs on port 5000. Run `ngrok http 5000` and take note of the forwarding
+URL (this means copy it).
```
rwblickhan: I think you could just say "copy the forwarding URL" here, instead of the parenthesized aside.

docs/ServingTheServer.md
```diff
@@ -0,0 +1,138 @@
...
+
+## 4: Configure Your Environment
+
+We use environmental variables to make everything go smoothly. Therefore, you
```
rwblickhan: Nit: "environment variables" not "environmental".

docs/ServingTheServer.md
```diff
@@ -0,0 +1,138 @@
...
+## 5: Running the App
+
+You are now ready to start the app! To start it, run `pipenv run launch`. This
+will automatically include your `.env` file. Be sure to already have DynamoDB up
+and running as well.
+
+## 6: Configure Slack
+
+Back to slack. Go [here][slack-events] and click "Enable Events". You will be
+asked to submit a request URL. Remember that URL that you memerized? Paste the
```
rwblickhan: Nit: "memorized".

docs/ServingTheServer.md
```diff
@@ -0,0 +1,138 @@
...
+that too.
+
+> **Important**: Please use `https` and refrain from placing an extra `/` at the
+> end of the URL. It messed me up.
+
+## 7: Actually Testing it
+
+This is the final and most important part: finally testing if it works or not.
+Go to your slack workspace and add your slack bot to the channel if you have yet
+to do so (just type `@<slackbot name>` and slack will ask if you want to invite
```
rwblickhan: You might want to say `@<bot name>` here, because technically "Slackbot" is the name of the built-in Slack bot, so this might confuse people.

cheukyin699:
A lot of things (e.g. returning instead of logging) should be addressed in #134, and thus this is blocked until it gets merged.

factory/__init__.py
cheukyin699: Yeah. It always messes me up a bit. Put that in as an issue for refactoring?

rwblickhan: Yep, go ahead. 👍

docs/ServingTheServer.md
cheukyin699: Must've been my brain on memes....

model/user.py
```diff
@@ -17,6 +17,26 @@ def __init__(self, slack_id):
         self.__image_url = ""
         self.__permissions_level = Permissions.member
 
+    def get_attachment(self):
+        """Return slack-formatted attachment (dictionary) for user."""
+        text_pairs = [
+            ('Slack ID', self.__slack_id),
+            ('Name', self.__name),
```
michaelenglo: Maybe we should have a separate this into its own class, perhaps call it attachment generator class or something

cheukyin699: comment added

cheukyin699:
codecov is getting annoying. Imma gonna close and reopen

PR #134, opened 2018-12-13
Core error handling with logs

### LINKED ISSUE #69: Handling errors at the core level
Currently `core.py` returns 0, 1, or -1 depending on whether or not a command follows a valid format, the command is successful, or the command results in an error. This should be changed so that error handling is done rather than returning a value that indicates an error took place.

rmcreyes:
# Pull Request

## Description

Used logs and direct messages to the user to handle errors.

## Testing

## Ticket(s)

Closes #69

### DIFF
--- a/command/core.py
+++ b/command/core.py
@@ -43,6 +43,6 @@ class Core:
         welcome = 'Welcome to Lauchpad!'
         try:
             self.__bot.send_dm(welcome, new_user_id)
-            return True
+            logging.info("user " + new_user_id + " added to database - user notified")
         except SlackAPIError:
-            return False
+            logging.error("user " + new_user_id + " added to database - user not notified")
--- a/command/core.py
+++ b/command/core.py
@@ -37,12 +37,12 @@ class Core:
 
     def handle_team_join(self, event_data):
         """Handle the event of a new user joining the workspace."""
-        new_user_id = event_data["event"]["user"]["id"]
-        new_user = User(new_user_id)
+        new_id = event_data["event"]["user"]["id"]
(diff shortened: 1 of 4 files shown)


command/core.py
```diff
@@ -20,25 +21,28 @@ def handle_app_mention(self, event_data):
...
         self.__facade.store_user(new_user)
         welcome = 'Welcome to Lauchpad!'
         try:
-            self.__bot.send_dm(welcome, new_user_id)
-            return True
+            self.__bot.send_dm(welcome, new_id)
+            logging.info(new_id + " added to database - user notified")
         except SlackAPIError:
-            return False
+            logging.error(new_id + " added to database - user not notified")
```
bobheadxi: just a thought, but have you guys thought about using [structured logging](https://www.structlog.org/en/stable/why.html) of some sort?  ie instead of logging:

```
time - mymessage, somecontext, somethinghappened
```

log:

```
{time: time, msg: mymessage, context: mycontext, userid: userid, error: error, ....}
```

this would allow us to run logs through some kind of log analysis in the future, and make things easier to query for debugging purposes, and allow more context without making log messages crazy long

for example, earlier (`logging.error("app mention event triggered incorrectly")`) instead of:

```
time: app mention event triggered incorrectly
```

we could get:

```
{time: time, msg: "app mention event triggered incorrectly", received_args: [arg1,arg2,...]}
```

rmcreyes: Yeah sure that sounds good! Should I make that my next ticket?

rwblickhan: Sure, that sounds like a good idea.

PR #140, opened 2018-12-23
Jacqueschen1/#130user command

### LINKED ISSUE #110: Finish user edit command implementation
Once database update function is added, we can implement the user edit command helper

jacqueschen1:
# Pull Request

## Description

Finished implementation of user edit command and tests

## Testing

## Ticket(s)

Closes #110 

### DIFF
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -138,17 +138,17 @@ class UserCommand:
             except LookupError:
                 return self.bot.send_to_channel(self.lookup_error, channel)
 
-        if param_list["name"] is not None:
+        if param_list["name"]:
             edited_user.set_name(param_list["name"])
-        if param_list["email"] is not None:
+        if param_list["email"]:
             edited_user.set_email(param_list["email"])
-        if param_list["pos"] is not None:
+        if param_list["pos"]:
             edited_user.set_position(param_list["pos"])
-        if param_list["github"] is not None:
+        if param_list["github"]:
             edited_user.set_github_username(param_list["github"])
-        if param_list["major"] is not None:
+        if param_list["major"]:
             edited_user.set_major(param_list["major"])
-        if param_list["bio"] is not None:
+        if param_list["bio"]:
             edited_user.set_biography(param_list["bio"])
 
         self.facade.store_user(edited_user)
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -96,8 +96,15 @@ class UserCommand:
             self.delete_helper(user_id, args.slack_id, channel)
 
         elif args.which == "edit":
-            param_list = [args.member, args.name, args.email,
-                          args.pos, args.github, args.major, args.bio]
+            param_list = {
+                "member": args.member,
+                "name": args.name,
+                "email": args.email,
+                "pos": args.pos,
+                "github": args.github,
+                "major": args.major,
+                "bio": args.bio,
+            }
             self.edit_helper(user_id, param_list, channel)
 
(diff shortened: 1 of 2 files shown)


command/commands/user.py
```diff
@@ -97,23 +96,58 @@ def handle(self, command, user_id, channel):
...
+        if param_list[0] is not None:
+            try:
+                admin_user = self.facade.retrieve_user(user_id)
+                if admin_user.get_permissions_level() != Permissions.admin:
+                    return self.bot.send_to_channel(self.permission_error,
+                                                    channel)
+                else:
+                    edited_id = param_list[0]
+                    edited_user = self.facade.retrieve_user(edited_id)
+                    self.facade.delete_user(edited_id)
```
cheukyin699: No need to delete user afterwards; the user is auto-updated when you store the user back in.

command/commands/user.py
```diff
@@ -97,23 +96,58 @@ def handle(self, command, user_id, channel):
...
+            except LookupError:
+                return self.bot.send_to_channel(self.lookup_error, channel)
+        else:
+            try:
+                edited_user = self.facade.retrieve_user(user_id)
+                self.facade.delete_user(user_id)
```
cheukyin699: Same as above

command/commands/user.py
```diff
@@ -97,23 +96,58 @@ def handle(self, command, user_id, channel):
...
+        if param_list[2] is not None:
+            edited_user.set_email(param_list[2])
+        if param_list[3] is not None:
+            edited_user.set_position(param_list[3])
+        if param_list[4] is not None:
+            edited_user.set_github_username(param_list[4])
+        if param_list[5] is not None:
+            edited_user.set_major(param_list[5])
+        if param_list[6] is not None:
+            edited_user.set_biography(param_list[6])
```
cheukyin699: Might be a bit arbitrary? Maybe try putting everything in a dictionary, or passing the original arguments so that you get the things named.

jacqueschen1: Yah doing it as a dictionary works too

command/commands/user.py
```diff
@@ -97,23 +96,63 @@ def handle(self, command, user_id, channel):
...
+        :return: returns error message if not admin and command
+                   edits another user, returns edit message if user is edited
+        """
+        edited_user = None
+        if param_list["member"] is not None:
+            try:
+                admin_user = self.facade.retrieve_user(user_id)
+                if admin_user.get_permissions_level() != Permissions.admin:
+                    return self.bot.send_to_channel(self.permission_error,
+                                                    channel)
```
cheukyin699: Don't know if this is intentional, but `bot.send_to_channel` does not return anything; you will be returning a `None` value instead.

command/commands/user.py
```diff
@@ -97,23 +96,63 @@ def handle(self, command, user_id, channel):
...
+                    edited_user = self.facade.retrieve_user(edited_id)
+            except LookupError:
+                return self.bot.send_to_channel(self.lookup_error, channel)
+        else:
+            try:
+                edited_user = self.facade.retrieve_user(user_id)
+            except LookupError:
+                return self.bot.send_to_channel(self.lookup_error, channel)
+
+        if param_list["name"] is not None:
```
cheukyin699: A more succinct way would be:

```python
if param_list["name"]:
    # ...
```

command/commands/user.py
jacqueschen1: I put that there so that the method would stop running there, idk if there's a better way of doing that though

PR #154, opened 2019-01-12
Slash commands

### LINKED ISSUE #147: Change to use slash commands instead of parsing dms
This makes it so that you can do the commands everywhere. Also makes a lot of the parsing and other things easier because it is the slack-sanctioned way of doing things.

This includes major refactoring.

cheukyin699:
# Pull Request

## Description

Refactor project to switch bases from listening to app_mention events, to handling slash command callbacks. Documentation is also changed. All irrelevant tests are removed and relevant tests are altered.

## Ticket(s)

Closes #147


command/commands/team.py
```diff
@@ -106,26 +106,32 @@ def get_help(self):
...
         """Handle command by splitting into substrings and giving to parser."""
         logging.debug("Handling TeamCommand")
         command_arg = shlex.split(command)
-        args = self.parser.parse_args(command_arg)
+        args = None
+
+        try:
+            args = self.parser.parse_args(command_arg)
+        except SystemExit:
+            return self.help, 200
```
rwblickhan: Maybe at some point we want to have "standardized" codes in an enum or something - obviously we would use 200 here, but in other places it's possible the return code is ambiguous. Not really important for now though.

server/server.py
```diff
@@ -35,6 +35,14 @@ def check():
     return "🚀"
 
 
+@app.route('/slack/commands', methods=['POST'])
```
rwblickhan: Can anything else be removed here (i.e. the Slack events adapter)? I'm not actually sure, and if it's risky we can just keep it for now and remove it in another PR.

PR #168, opened 2019-01-16
DynamoDB switch to Github ID

### LINKED ISSUE #155: Switch slack IDs in team table to Github ID
Currently the users and teams tables are strongly coupled, as the team table uses the slack IDs that are the primary keys of the users table. However, since we will assume Github is the source of truth for teams, we should instead store Github IDs.

### LINKED ISSUE #156: Switch teams table primary key to team ID
Github has a concept of team ID, which we should use instead of the team name, which can change.

cheukyin699:
# Pull Request

## Description

Change DynamoDB teams table to use Github ID instead of Github username or Github team names.

## Ticket(s)

Closes #155, closes #156.


docs/Database.md
```diff
@@ -28,7 +29,8 @@ and are also listed here:
 
 Attribute Name | Description
 ---|---
-`github_team_name` | `String`; The team's github name
+`gtid` | `String`; The team's Github ID
```
rwblickhan: Is `gtid` used in the official Github documentation? If not, I think I would rather have this as `github_team_id` or something similar - `gtid` doesn't seem descriptive enough to me.

db/dynamodb.py
```diff
@@ -209,7 +210,7 @@ def retrieve_user(self, slack_id):
         """
         user_table = self.ddb.Table(self.users_table)
         response = user_table.get_item(
-            TableName='users',
+            TableName=self.users_table,
```
rwblickhan: This is also fixed by #167.

db/dynamodb.py
```diff
@@ -250,15 +251,16 @@ def retrieve_team(self, team_name):
         """
         team_table = self.ddb.Table(self.teams_table)
         response = team_table.get_item(
-            TableName='teams',
+            TableName=self.teams_table,
```
rwblickhan: Also fixed in #167.

model/team.py
```diff
@@ -21,11 +23,13 @@ def is_valid(team):
 
         Required fields for database to accept:
         - ``__github_team_name``
+        - ``__gtid``
```
rwblickhan: It _might_ be a good idea to make these part of a two-part key in DynamoDb, if both are necessary. It's not strictly necessary but it would keep everything consistent.

model/team.py
```diff
@@ -55,21 +67,21 @@ def get_platform(self):
         """Return this team's working platform."""
         return self.__platform
 
-    def add_member(self, slack_id):
-        """Add a new member's Slack ID to the team's set of members' IDs."""
-        self.__members.add(slack_id)
+    def add_member(self, gid):
```
rwblickhan: Inconsistent - it would probably be better if these parameters are also name `guid` (although I don't like the name `guid` either).

rwblickhan:
Looks like Travis is failing because line 75 of `team.py` is too long.

docs/Database.md
```diff
@@ -28,7 +29,8 @@ and are also listed here:
 
 Attribute Name | Description
 ---|---
-`github_team_name` | `String`; The team's github name
+`gh_team_id` | `String`; The team's Github ID
+`github_team_name` | `String`; The team's Github name
```
rwblickhan: I don't like the inconsistency; we have:
* `gh_team_id`
* `github_team_name`
* `guid`
I'd prefer if these all followed a set naming convention (and, if possible, one that included the Slack ID as well).

cheukyin699: I think leaving slack ID labelled `slack_id` should be enough, at least for the length. Naming it `s_id` would be a bit confusing. I have gone and renamed all instances of `github...` and `g...` to `gh...`.

cheukyin699: nvm; renamed all instances to `github_`

Issue #161, opened 2019-01-12
Create organization Github API interface


rwblickhan:

rwblickhan:
Closed by #170.

PR #170, opened 2019-01-19
GitHub api organization

jacqueschen1:
# Pull Request

## Description
Adds GithubInterface interface, which interacts with pyGithub. Added methods for organization.

## Testing

## Ticket(s)

Affects #161 

### DIFF
--- a/command/core.py
+++ b/command/core.py
@@ -29,7 +29,7 @@ class Core:
         new_id = event_data["event"]["user"]["id"]
         new_user = User(new_id)
         self.__facade.store_user(new_user)
-        welcome = 'Welcome to Launch Pad!'
+        welcome = 'Welcome to UBC Launch Pad!'
         try:
             self.__bot.send_dm(welcome, new_id)
             logging.info(new_id + " added to database - user notified")
--- a/command/core.py
+++ b/command/core.py
@@ -29,7 +29,7 @@ class Core:
         new_id = event_data["event"]["user"]["id"]
         new_user = User(new_id)
         self.__facade.store_user(new_user)
-        welcome = 'Welcome to Lauchpad!'
+        welcome = 'Welcome to Launchpad!'
         try:
             self.__bot.send_dm(welcome, new_id)
             logging.info(new_id + " added to database - user notified")
--- a/command/core.py
+++ b/command/core.py
@@ -29,7 +29,7 @@ class Core:
         new_id = event_data["event"]["user"]["id"]
         new_user = User(new_id)
         self.__facade.store_user(new_user)
-        welcome = 'Welcome to Launchpad!'
+        welcome = 'Welcome to Launch Pad!'
         try:
             self.__bot.send_dm(welcome, new_id)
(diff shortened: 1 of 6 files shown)


cheukyin699:
Turns out that the issue is that you are importing the **MODULE** `github.Organization` instead of using `github.Organization.Organization`. What the Frick.

rwblickhan:
Looks like Travis is failing because line 23 of `github_test` is too long.

command/core.py
```diff
@@ -29,7 +29,7 @@ def handle_team_join(self, event_data):
         new_id = event_data["event"]["user"]["id"]
         new_user = User(new_id)
         self.__facade.store_user(new_user)
-        welcome = 'Welcome to Lauchpad!'
+        welcome = 'Welcome to Launchpad!'
```
rwblickhan: Launch Pad 😛

rwblickhan: Actually, we might even want to say "UBC Launch Pad"... we semi-consistently use the full name.

interface/github.py
```diff
@@ -0,0 +1,58 @@
+"""Utility classes for interacting with Github API via PyGithub."""
+from github import Github, GithubException
+
+
+class GitBot:
```
rwblickhan: I'm not really a fan of the name `GitBot` - this doesn't really have anything to do with being a bot (even though it is consistent with `SlackBot`). I'd prefer both this and the Slack class get renamed - honestly, even `Slack` and `Github` might be okay, assuming they're in the `interface` module.

interface/github.py
```diff
@@ -0,0 +1,58 @@
...
+        try:
+            self.github = github
+            self.org = self.github.get_organization(org)
+        except GithubException as e:
+            raise GithubAPIException(e.data)
+
+    def org_add_member(self, username):
+        """Add/update to member with given username to organization."""
+        try:
+            user = self.github.get_user(username)
```
rwblickhan: I assume this is retrieving the Github ID? It might be better to have this method (and all the other methods) take Github ID, since we're storing Github ID and treating that as the "canonical" representation of a user with regards to Github.

command/core.py
jacqueschen1: Oops ok

interface/github.py
jacqueschen1: haha ok

interface/github.py
jacqueschen1: mmm ill check the dots to see if that works

interface/github.py
jacqueschen1: Naming it github might be confusing since the pygithub class is also called github, is GitInterface ok?

interface/github.py
```diff
@@ -0,0 +1,58 @@
+"""Utility classes for interacting with Github API via PyGithub."""
+from github import Github, GithubException
+
+
+class GithubInterface:
+    """Utility class for interacting with Github API."""
```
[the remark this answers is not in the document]
jacqueschen1: Will add in the next pr

PR #174, opened 2019-01-20
Update local development guide

### LINKED ISSUE #145: Update local development docs to include containerization
The local development tutorial is written assuming you're running the server directly. We should update it to explain how to run it via Docker.

Also, we might want to rename those docs... Currently they're titled "Linking Your Dev Server to Slack", which is correct, but it might be more obvious to call them "Local Development Tutorial" or something similar.

### LINKED ISSUE #165: Document using AWS DynamoDB for manual testing instead of local DynamoDB
Please.

### LINKED ISSUE #173: Update local testing docs to include new config
Our local development docs still suggest using the local instance of DynamoDb and don't explain how to use the new configuration files. These should be updated.

rwblickhan:
# Pull Request

## Description

Updates the documentation.

## Ticket(s)

Closes #145
Closes #165
Closes #173


docs/LocalDevelopmentGuide.md
```diff
@@ -0,0 +1,223 @@
...
+to set it up.
+
+## 3: Configure a Slack App
+
+Follow the link [here][make-slack-app] to create a new Slack app - you can name it
+whatever you like - and install it to the appropriate workspace.
+
+### Add a Bot User
+
+In "Add features and functionality," add a bot user. Since this is just for testing,
```
omar2535: Styling mistake?

docs/LocalDevelopmentGuide.md
```diff
@@ -0,0 +1,223 @@
...
+you can name the bot user whatever you like.
+
+### Add Event Subscriptions
+
+In "Add features and functionality," add event subscriptions. In particular, under
```
omar2535: Same here

docs/LocalDevelopmentGuide.md
rwblickhan: See, I was taught to put the punctuation _inside_ the quotes... 😛

I can change it.

docs/LocalDevelopmentGuide.md
```diff
@@ -0,0 +1,223 @@
...
+whatever you like - and install it to the appropriate workspace.
+
+### Add a Bot User
+
+In "Add features and functionality", add a bot user. Since this is just for testing,
+you can name the bot user whatever you like.
+
+### Add Event Subscriptions
+
+In "Add features and functionality", add event subscriptions. In particular, under
```
omar2535: I suggest moving this after step 6: adding in credentials for `slack.toml`. Reason is because when you add a event subscription, it checks against the forwarded ngrok url at /slack/events but will get an error 500 since the app isn't running yet - thus blocking you from making the feature for the bot (as in you wont be able to save changes on the slack app).

cheukyin699:
@rwblickhan Yes it does.

cheukyin699:
Edited to reflect this.

PR #181, opened 2019-01-26
Update documentation

### LINKED ISSUE #171: Fix documentation on slash commands, and mandatory Github IDs
Since things change, a portion of the docs weren't up to date (there are portions that still use @ commands and not use Github IDs). Thus, this needs to be fixed immediately. We must first wait until #168 gets merged in order to regenerate the API docs. Then we can change the instances of `@rocket` to `/rocket`.

The file `docs/GettingStarted.md` contains some out-of-date code snippets about the database and the models that need to be updated. Please do so.

cheukyin699:
# Pull Request

## Description

Updates the documentation

## Testing

You can view the docs [here][there].

## Ticket(s)

Closes #171.

[there]: https://rocket2.readthedocs.io/en/cheukyin699--171-docs-fix/

### DIFF
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -168,8 +168,9 @@ class UserCommand:
         Delete user from database.
 
         Delete user with ``slack_id`` from database if user with ``user_id``
-        has admin permission level. **Note**: deleting yourself is entirely
-        possible.
+        has admin permission level.
+
+        **Note**: user can delete themselves.
 
         :param user_id: Slack ID of user who is calling the command
         :param slack_id: Slack ID of user who is being deleted
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -121,8 +121,9 @@ class UserCommand:
         """
         Edit user from database.
 
-        If param_list[0] is not None, edits user with
-        ID param_list[0], else edits user with user_id
+        If ``param_list['member']`` is not ``None``, this function edits using
+        the ID from ``param_list['member']`` (must be an admin to do so).
+        Otherwise, edits the user that called the function.
 
         :param user_id: Slack ID of user who is calling the command
         :param param_list: List of user parameters that are to be edited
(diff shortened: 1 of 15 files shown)


docs/Testing.md
```diff
@@ -29,17 +33,15 @@ get everything working.
...
-`TESTING=false pytest`
+This is the recommended way for unit testing.
 
 ### Run server DynamoDB
 
-To run the server DynamoDB we need to set the `REGION`. The `REGION` is default
-to `us-east-1`, however you can change it to [any of the regions
-here](https://docs.aws.amazon.com/general/latest/gr/rande.html).
+To run the server DynamoDB we need to set the `region` and obtain both the
+AWS `access_key_id` and `secret_access_key`.
```
rwblickhan: May want to include a pointer to the full local development setup documentation that explains this.

command/commands/user.py
```diff
@@ -166,13 +167,14 @@ def delete_helper(self, user_id, slack_id):
         """
         Delete user from database.
 
-        Delete user with slack_id from database if user with user_id has
-        admin permission level.
+        Delete user with ``slack_id`` from database if user with ``user_id``
+        has admin permission level. **Note**: deleting yourself is entirely
```
michaelenglo: Not sure why you would ever want to delete yourself. Can you tell me a specific case of where this is useful?

tests/util.py
```diff
@@ -6,7 +6,21 @@
 
 def create_test_user(slack_id):
```
michaelenglo: Perhaps, we can be more explicit here and rename the method `create_test_admin_user` since admin user is not the best representation of all users. What do you think?

tests/util.py
```diff
@@ -6,7 +6,21 @@
...
+
+    ==========  =============================
+    Property    Preset
+    ==========  =============================
+    Slack ID    ``slack_id``
+    Bio         I like puppies and kittens!
+    Email       admin@ubc.ca
+    Name        Iemann Atmin
+    Github      kibbles
+    Image URL   https://google.ca
```
michaelenglo: nitpick: maybe find an actual image URL, not just any URL. Might be useful if we actually want to test future modules that require retrieving an actual image.

command/commands/user.py
rwblickhan: I think that's the point of his comment - it's rather unintuitive to expect to be able to delete yourself, yet because of the way we've designed things it is actually possible, so be careful not to do it. (Though in that case, I agree - this could be reworded slightly, to sound more like a warning rather than just a random parenthetical comment.)

tests/util.py
cheukyin699: Yes

PR #182, opened 2019-01-30
set up server.py to receive github org webhook

### LINKED ISSUE #175: Hook up webhook handler in server.py

rmcreyes:
# Pull Request

## Description

Added a route to `server/server.py` so that rocket receives organization webhook events from GitHub. Also refactored `handle_organization_event()` so it expects a dictionary as an argument rather than a JSON string. Added a factory method to `factory/__init__.py` to align with the construction of `Core`, but I'm open to suggestions for changing this.

## Testing

This was tested by creating a test organization and adding a webhook to its settings, where the payload URL is specified in `server.py`. I manually tested using a fake github account leaving and entering the organization.

## Ticket(s)

Closes #175 

### DIFF
--- a/tests/webhook/webhook_test.py
+++ b/tests/webhook/webhook_test.py
@@ -1,5 +1,4 @@
 """Test the webhook handler."""
-import json
 import pytest
 from unittest import mock
 from model.user import User
--- a/tests/webhook/webhook_test.py
+++ b/tests/webhook/webhook_test.py
@@ -8,56 +8,20 @@ from db.facade import DBFacade
 
 
 @pytest.fixture
-def org_default_payload_string():
+def org_default_payload():
     """Provide the basic structure for an organization payload."""
     default_payload =\
-        """
-            {
-                "action": "member_added",
-                "membership": {
-                    "url": "",
-                    "state": "pending",
-                    "role": "member",
-                    "organization_url": "",
-                    "user": {
-                        "login": "hacktocat",
-                        "id": 39652351,
-                        "node_id": "MDQ6VXNlcjM5NjUyMzUx",
-                        "avatar_url": "",
-                        "gravatar_id": "",
-                        "url": "",
-                        "html_url": "",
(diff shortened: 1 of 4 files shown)


tests/webhook/webhook_test.py
```diff
@@ -141,10 +142,12 @@ def test_handle_org_event_rm_member_missing(mock_logging, org_rm_payload):
     mock_facade = mock.MagicMock(DBFacade)
     mock_facade.query_user.return_value = []
     webhook_handler = WebhookHandler(mock_facade)
-    webhook_handler.handle_organization_event(org_rm_payload)
+    rsp, code = webhook_handler.handle_organization_event(org_rm_payload)
     mock_facade.query_user\
-        .assert_called_once_with(['github_id', 39652351])
+        .assert_called_once_with([('github_id', 39652351)])
```
michaelenglo: wait why facade returns an array of tuples now? I thought query user only queries one user at the moment

tests/webhook/webhook_test.py
```diff
@@ -156,30 +159,36 @@ def test_handle_org_event_rm_mult_members(mock_logging, org_rm_payload):
     user3 = User("SLACKUSER3")
     mock_facade.query_user.return_value = [user1, user2, user3]
     webhook_handler = WebhookHandler(mock_facade)
-    webhook_handler.handle_organization_event(org_rm_payload)
+    rsp, code = webhook_handler.handle_organization_event(org_rm_payload)
     mock_facade.query_user\
-        .assert_called_once_with(['github_id', 39652351])
+        .assert_called_once_with([('github_id', 39652351)])
```
michaelenglo: This too

tests/webhook/webhook_test.py
rmcreyes: It doesn't, `query_user()` returns a list of users given a list of tuples, where each tuple is an attribute and value

PR #194, opened 2019-02-03
Add user to organization when username is edited

### LINKED ISSUE #176: Add user to organization when Github username is edited
When a user edits their Github username (i.e. via the `/rocket user edit` command via `UserCommand`), we will need to make a request to the Github organization API interface created in #170 to add that user's Github account to the appropriate Github organization.

A few things to keep in mind:
* The PR for #170 doesn't yet construct an instance of the Github API interface, which should most likely be done in `core.py`. Talk to @jacqueschen1 to see if he'll still do this or if you can take it over.
* The Github interface requires an organization when launched. This should be configurable, so add a field to `config.toml` (possibly in a new `github` section?) that is then read from before constructing an instance of the Github interface.

ravenCondol:
# Pull Request

## Description

The command `/rocket user edit --github 'newusername'` should now be making a request to the Github organization API interface created in #170 to add newly provided github username to the appropriate organization. 

Will have to add documentation about new config file github.toml, needed for the api key and the organization name (which should be ubclaunchpad ?)

## Testing

## Ticket(s)

Closes #176

### DIFF
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -157,7 +157,7 @@ class UserCommand:
         if param_list["github"]:
             edited_user.set_github_username(param_list["github"])
             try:
-                github.org_add_username(param_list["github"])
+                self.github.org_add_username(param_list["github"])
             except Exception as e:
                 logging.error("Error while editing user organization!\n" + e)
         if param_list["major"]:
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -5,6 +5,7 @@ import shlex
 from flask import jsonify
 from model.permissions import Permissions
 from model.user import User
+from interface.github import GithubAPIException
 
 
 class UserCommand:
@@ -31,14 +32,14 @@ class UserCommand:
     lookup_error = "User not found!"
     delete_text = "Deleted user with Slack ID: "
 
-    def __init__(self, db_facade, githubInterface):
+    def __init__(self, db_facade, github_interface):
         """Initialize user command."""
         logging.info("Initializing UserCommand instance")
         self.parser = argparse.ArgumentParser(prog="user")
         self.parser.add_argument("user")
         self.init_subparsers()
         self.facade = db_facade
-        self.github = githubInterface
+        self.github = github_interface
 
     def init_subparsers(self):
         """Initialize subparsers for user command."""
@@ -158,8 +159,8 @@ class UserCommand:
             edited_user.set_github_username(param_list["github"])
             try:
                 self.github.org_add_member(param_list["github"])
-            except Exception as e:
-                logging.error("Error while editing user organization!\n")
+            except GithubAPIException as e:
+                logging.error("Error while editing user's github organization!\n")
         if param_list["major"]:
             edited_user.set_major(param_list["major"])
         if param_list["bio"]:
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -5,7 +5,6 @@ import shlex
 from flask import jsonify
 from model.permissions import Permissions
 from model.user import User
-from command.core import Core
 
 
 class UserCommand:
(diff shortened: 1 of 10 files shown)


command/commands/user.py
```diff
@@ -31,13 +31,14 @@ class UserCommand:
     lookup_error = "User not found!"
     delete_text = "Deleted user with Slack ID: "
 
-    def __init__(self, db_facade):
+    def __init__(self, db_facade, githubInterface):
```
rwblickhan: Please use `snake_case` for consistency (so in this case `github_interface`).

command/core.py
```diff
@@ -8,12 +8,13 @@
 class Core:
     """Encapsulate methods for handling events."""
 
-    def __init__(self, db_facade, bot):
+    def __init__(self, db_facade, bot, ghInterface):
```
rwblickhan: Use `snake_case` for consistency.

credentials/github.toml
```diff
@@ -0,0 +1,2 @@
+api_token = 
+organization = ubclaunchpad
```
rwblickhan: `organization` isn't really credentials - I would put it in the main config (the `ubclaunchpad` default should be fine though, yeah).

credentials/github.toml
```diff
@@ -0,0 +1,2 @@
+api_token =
```
rwblickhan: Don't commit this file to the repo - in fact, I'm pretty sure this entire folder should be in `.gitignore`? If not, it should be added.

factory/__init__.py
```diff
@@ -5,24 +5,30 @@
...
     Initialize and returns a :class:`command.core.Core` object.
 
     :return: a new ``Core`` object, freshly initialized
     """
+    slack_api_token, github_api_token, github_organization = "", "", ""
     if not config['testing']:
-        slack_api_token = toml.load(config['slack']['creds_path'])['api_token']
-    else:
-        slack_api_token = ""
+        slack_api_token += toml.load(
```
rwblickhan: Out of curiosity, is there a reason you used `+=`? (I don't think it's a problem, just curious about the reason for the change.)

command/commands/user.py
```diff
@@ -155,6 +156,10 @@ def edit_helper(self, user_id, param_list):
             edited_user.set_position(param_list["pos"])
         if param_list["github"]:
             edited_user.set_github_username(param_list["github"])
+            try:
+                self.github.org_add_member(param_list["github"])
+            except Exception as e:
```
rwblickhan: Catching `Exception` is probably unnecessarily broad - I would catch only the `GithubAPIException` exposed by the `interface` module.

command/commands/user.py
```diff
@@ -31,13 +31,14 @@ class UserCommand:
     lookup_error = "User not found!"
```
rwblickhan: Ignore this...

credentials/github.toml
rwblickhan: Oh, and also, although I believe this works, I think it's "more proper" TOML to wrap the field (i.e. `ubclaunchpad`) in quotes.

credentials/github.toml
ravenCondol: Since it was the first time this file was made, I thought it would need to be added at least once. How do i remove this file?

rwblickhan: Ah, makes sense. Yeah, we don't want it to be added, just in case somebody accidentally commits credentials. If you remove the file, git should have a "file deleted" change - if you commit that, it should be removed from the repo (and, since I did check and this directory is in `.gitignore`, it shouldn't be added again even if you change it later).

factory/__init__.py
ravenCondol: I have no idea why I did, I can remove if needed?

command/commands/user.py
```diff
@@ -155,6 +157,10 @@ def edit_helper(self, user_id, param_list):
             edited_user.set_position(param_list["pos"])
         if param_list["github"]:
             edited_user.set_github_username(param_list["github"])
+            try:
+                self.github.org_add_member(param_list["github"])
+            except GithubAPIException as e:
+                logging.error("Error editing github organization!\n")
```
cheukyin699: I would change the error message to:

```python
logging.error("Error adding user %s to GitHub organization" % param_list['github'])
```

Also, we should tell the user about the error. It should be appended to the `User edited` thing as a new line or something. Just for some feedback to the user.

Also, you may want to place line 159 after line 161 since we may not want to add a GitHub username to the model if we can't even add it to the GitHub organization.

PR #196, opened 2019-02-03
Fix Apple smart punctuation bug

### LINKED ISSUE #195: Parser does not handle ’ as '
We just found out that Apple devices would turn your single apostrophes into their Unicode counterparts to look better. Unfortunately for everyone involved, they aren't parsed with `argparse`, and are treated as just your average Unicode character instead.

We want to replace all ’ and their friends (the Unicode starting/ending double-quotes, Unicode starting/ending single-quotes) with their counterparts. This should happen when the command comes in as a string, but before the command parsing takes place.

rwblickhan:
# Pull Request

## Description

Fixes the Apple smart punctuation bug described in #195, albeit in a slightly hacky way.

## Ticket(s)

Closes #195

### DIFF
--- a/command/core.py
+++ b/command/core.py
@@ -17,6 +17,16 @@ class Core:
 
     def handle_app_command(self, cmd_txt, user):
         """Handle a command call to rocket."""
+        def regularize_char(c):
+            if c == "‘" or c == "’":
+                return "'"
+            if c == '“' or c == '”':
+                return '"'
+            return c
+
+        # Slightly hacky way to deal with Apple platform
+        # smart punctuation messing with argparse.
+        cmd_txt = ''.join(map(regularize_char, cmd_txt))
         s = cmd_txt.split(' ', 1)
         if s[0] in self.__commands:
             return self.__commands[s[0]].handle(cmd_txt, user)


ravenCondol:
Not about this, but I notice that the travis push build failed because of some db tests. This happened to me, too, and resolves most of the time by rebuilding, but like why does it sometimes happen at all?

rwblickhan:
I'm... not really sure, to be honest. We should probably look into this at some point. Sometimes tests, especially complicated tests like this where we have a full DynamoDb instance running, can be flaky.

command/core.py
```diff
@@ -17,6 +17,16 @@ def __init__(self, db_facade, bot):
...
+        def regularize_char(c):
+            if c == "‘" or c == "’":
+                return "'"
+            if c == '“' or c == '”':
+                return '"'
+            return c
+
+        # Slightly hacky way to deal with Apple platform
+        # smart punctuation messing with argparse.
+        cmd_txt = ''.join(map(regularize_char, cmd_txt))
```
cheukyin699: <3<3

cheukyin699:
It's best not to. **SOMETIMES THERE MAY BE PEOPLE WHO PUT UNSAFE QUOTES INSIDE STRINGS, AND WE SHOULD GO EASY ON THEM**

rwblickhan:
@cheukyin699 seems upset about people's choice of quotes 🤔

cheukyin699:
There are times where we can let things pass. Then there are other times where we may have to consider all the Youth in Asia.

PR #200, opened 2019-02-06
Fix #197 - user add does not overwrite by default

### LINKED ISSUE #197: /rocket user add overwrites existing user
It seems like `/rocket user add`, when applied to a user that already exists in the Rocket database, will overwrite their entry with an empty one, without a prompt. We should reconsider this behaviour - perhaps `/rocket user add -f` to force the overwrite when that's actually the intended behaviour?

cheukyin699:
# Pull Request

## Description

Add command option to force overwriting of user entry in database, and disable overwriting by default by checking to see if the user exists.

## Ticket(s)

Closes #197.

### DIFF
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -51,9 +51,9 @@ class UserCommand:
         parser_view.add_argument("--slack_id", type=str, action='store')
 
         """Parser for add command."""
-        # DEBUG
         parser_add = subparsers.add_parser("add")
         parser_add.set_defaults(which="add")
+        parser_add.add_argument("-f", "--force", action="store_true")
 
         """Parser for help command."""
         parser_help = subparsers.add_parser("help")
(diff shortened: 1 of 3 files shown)


command/commands/user.py
```diff
@@ -218,12 +218,22 @@ def view_helper(self, user_id, slack_id):
...
 
         :param user_id: Slack ID of user to be added
+        :param use_force: If this is set, we store the user even if they are
+        already added in the database
         :return: ``"User added!", 200``
         """
+        # Try to look up and avoid overwriting if we are not using force
+        if not use_force:
+            try:
+                self.facade.retrieve_user(user_id)
```
rwblickhan: It might be a good idea to return a message here, maybe something along the lines of "that user already exists; to overwrite with a new user use -f".

Issue #202, opened 2019-02-08
Refactor individual command help text delivery


rwblickhan:
In #193 @bobheadxi pointed out the [help printing](https://docs.python.org/3/library/argparse.html#printing-help) utilities in `argparse`. Currently, we don't use these, but instead use a hardcoded help string for each command. We should refactor to make use of `format_usage` and/or `format_help` instead of a hardcoded help strings.

Issue #65, opened 2018-10-29
Research metaclasses for commands framework


rwblickhan:
It looks like we'll want to use metaclasses to enforce an interface for the commands framework and possibly create the commands on launch. We should research this more.

rwblickhan:
I suspect we'll simply want an [abstract base class](https://docs.python.org/3.7/library/abc.html) instead - metaclasses are probably overkill.

PR #204, opened 2019-02-08
Added a means for core to display all help messages of its commands. …

### LINKED ISSUE #193: Implement help command
When we type `/rocket help`, we want to display a help menu, somewhat similar to the original version of Rocket. This can probably be hardcoded into `handle_app_command()` in `Core`. In particular, each `Command` in `__commands` already has a help string available - the best implementation is probably to get the help string for each `Command`, concatenate them (with some formatting), then return it with a 200 status code.

ravenCondol:
# Pull Request

## Description

Provides bot behavior for "/rocket help", which should display a pretty simple help menu. Design can be changed easily enough if need be!

## Ticket(s)

Closes #193

### DIFF
--- a/command/core.py
+++ b/command/core.py
@@ -3,6 +3,7 @@ from command.commands.user import UserCommand
 from model.user import User
 from interface.slack import SlackAPIError
 import logging
+import json
 
 
 class Core:
@@ -29,6 +30,9 @@ class Core:
         # smart punctuation messing with argparse.
         cmd_txt = ''.join(map(regularize_char, cmd_txt))
         s = cmd_txt.split(' ', 1)
+        if s[0] == "help" or s[0] is None:
+            logging.info("Help command was called")
+            return self.get_help(), 200
         if s[0] in self.__commands:
             return self.__commands[s[0]].handle(cmd_txt, user)
         else:
@@ -46,3 +50,22 @@ class Core:
             logging.info(new_id + " added to database - user notified")
         except SlackAPIError:
             logging.error(new_id + " added to database - user not notified")
+
+    def get_help(self):
+        message = {"text": "Displaying all available commands. To read about"
+                           " a specific command, use `/rocket COMMAND help`\n",
+                   "mrkdwn": "true"}
+        attachments = []
+        for cmd in self.__commands:
+            cmd_name = self.__commands[cmd].get_name()
+            cmd_text = "*[" + cmd_name.upper() + "]*\n\n" + \
+                "Commands for editing", cmd_name + "s."
+            # use below when cmd_usage is fixed in #202
+            # cmd_usage = self.__commands[cmd].get_help()
+            # cmd_text = "*[" + cmd_name + "]*\n\n" + \
+            #           "```" + cmd_usage + "```"
+            # perhaps get_help could have a parameter to only return usage?
+            attachment = {"text": cmd_text, "mrkdwn_in": ["text"]}
+            attachments.append(attachment)
+        message["attachments"] = attachments
+        return json.dumps(message)
--- a/command/core.py
+++ b/command/core.py
@@ -53,19 +53,14 @@ class Core:
 
     def get_help(self):
         """Get help messages and return a formatted string for messaging."""
-        message = {"text": "Displaying all available commands. To read about"
-                           " a specific command, use `/rocket COMMAND help`\n",
+        message = {"text": "Displaying all available commands. "
+                           "To read about a specific command, use "
+                           "\n`/rocket [command] help`\n",
                    "mrkdwn": "true"}
         attachments = []
-        for cmd in self.__commands:
-            cmd_name = self.__commands[cmd].get_name()
-            cmd_text = "*[" + cmd_name.upper() + "]*\n\n" + \
-                "Commands for editing", cmd_name + "s."
-            # use below when cmd_usage is fixed in #202
-            # cmd_usage = self.__commands[cmd].get_help()
-            # cmd_text = "*[" + cmd_name + "]*\n\n" + \
-            #           "```" + cmd_usage + "```"
-            # perhaps get_help could have a parameter to only return usage?
+        for cmd in self.__commands.values():
+            cmd_name = cmd.get_name()
+            cmd_text = "*" + cmd_name + ":* " + cmd.get_desc()
             attachment = {"text": cmd_text, "mrkdwn_in": ["text"]}
             attachments.append(attachment)
         message["attachments"] = attachments
--- a/command/core.py
+++ b/command/core.py
@@ -3,7 +3,7 @@ from command.commands.user import UserCommand
 from model.user import User
 from interface.slack import SlackAPIError
 import logging
-import json
+from flask import jsonify
 
 
 class Core:
@@ -64,4 +64,4 @@ class Core:
             attachment = {"text": cmd_text, "mrkdwn_in": ["text"]}
             attachments.append(attachment)
         message["attachments"] = attachments
-        return json.dumps(message)
+        return jsonify(message)
--- a/command/core.py
+++ b/command/core.py
@@ -52,6 +52,7 @@ class Core:
             logging.error(new_id + " added to database - user not notified")
 
     def get_help(self):
+        """Get help messages and return a formatted string for messaging."""
         message = {"text": "Displaying all available commands. To read about"
                            " a specific command, use `/rocket COMMAND help`\n",
                    "mrkdwn": "true"}
--- a/tests/command/core_test.py
+++ b/tests/command/core_test.py
@@ -34,6 +34,7 @@ def test_handle_invalid_command(mock_logging, mock_usercommand):
     core.handle_app_command('fake command', user)
     error_txt = "Please enter a valid command."
 
+
 @mock.patch('command.core.logging')
 def test_handle_help(mock_logging,):
     """Test that a '/rocket help' brings up help"""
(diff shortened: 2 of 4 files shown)


rwblickhan:
Looks like there’s a couple style issues you’ll have to fix.

ravenCondol:
oh nooo, also why is it out of date

cheukyin699:
We just merged a PR. You can fetch, pull, rebase, and push again to update

ravenCondol:
... okay hold on

command/core.py
```diff
@@ -46,3 +50,23 @@ def handle_team_join(self, event_data):
...
+        """Get help messages and return a formatted string for messaging."""
+        message = {"text": "Displaying all available commands. To read about"
+                           " a specific command, use `/rocket COMMAND help`\n",
+                   "mrkdwn": "true"}
+        attachments = []
+        for cmd in self.__commands:
+            cmd_name = self.__commands[cmd].get_name()
+            cmd_text = "*[" + cmd_name.upper() + "]*\n\n" + \
+                "Commands for editing", cmd_name + "s."
+            # use below when cmd_usage is fixed in #202
```
rwblickhan: Hmm, I don't think it'll be a good idea to display the full usage for a command with `/rocket help`, even after #202 is fixed. What might be the best idea here is to add a hardcoded "description" field for each command, then you can concatenate that onto `cmd_name` and just ignore `cmd_usage` (because, as you suggest in the help text intro, users should just run `/rocket COMMAND help` to see usage for a particular command).

rwblickhan: In particular, that would lead to output like
```
user: interact with users
team: interact with teams
...
```
which I think would be more useful to users than a list of usage for every single subcommand.

command/core.py
```diff
@@ -46,3 +50,23 @@ def handle_team_join(self, event_data):
...
+    def get_help(self):
+        """Get help messages and return a formatted string for messaging."""
+        message = {"text": "Displaying all available commands. To read about"
+                           " a specific command, use `/rocket COMMAND help`\n",
+                   "mrkdwn": "true"}
+        attachments = []
+        for cmd in self.__commands:
+            cmd_name = self.__commands[cmd].get_name()
+            cmd_text = "*[" + cmd_name.upper() + "]*\n\n" + \
+                "Commands for editing", cmd_name + "s."
```
rwblickhan: Out of curiosity, what's the `"s."` for?

rwblickhan:
By the way, @ravenCondol, do you have pictures of the output handy? That might be useful for us as reviewers.

command/core.py
ravenCondol: It is to make the names plural, e.g. It would look like "commands for editing users, commands for editing teams

rwblickhan: Ooh I see, that's clever. One thing to keep in mind there, though, is that not all commands will necessarily follow that naming convention in the future (though it's probably fine for now).

command/core.py
ravenCondol: what is the difference between format_usage and format_help when printing out argparse instructions? i can't really make sense of which is used for what in the examples provided by the docs

ravenCondol:
@rwblickhan i do! its using slack's message builder, where i can paste the json string returned as an argument for the get request: https://api.slack.com/docs/messages/builder?msg={place json stuff with symbols escaped here} . there's probably a way to test this by using the bot itself but i still haven't gotten it set up yet ... i'll post pictures of the output here in a few minutes

command/core.py
rwblickhan: I'm _guessing_ `format_help` is what we want, though seeing as how I haven't used it before I'm honestly not sure myself 😛 We'll just have to play around with it a bit probably.

That can be done in #202 though, I think.

ravenCondol: wait so i dont need to change anything else? besides like posting pics of the output

ravenCondol:
image

command/core.py
rwblickhan: Oh, sorry, I think I'm being unclear 😂 Kinda confusing that we have two different levels of help implemented in completely different ways, and I think I also slightly misunderstood your code...

So what I'm getting at (now that I understand it all better 😂) is that you're hardcoding "Commands for editing *blank*". I think what might be a better/more extensible design is to have a `description` field for each command (which might just be "Commands for editing users" or whatever for now - I don't mind if it's hardcoded into the command itself) which is then concatenated on like you have now. We could then later on enforce that all commands actually have the `description` field, but they could have whatever description that makes sense for that command.

So, yep, all the `argparse` stuff isn't necessary to do yet - I would just prefer a `description` field in each command (with getters/setters/whatever) that then gets concatenated on similar to what you have now. You could also remove the note about adding it after #202 - the point here is that `get_help` or whatever we end up having for command-specific help doesn't need to interact with this at all (e.g. if we do `/rocket help`, we don't need to see all the options for `/rocket user`).

Anyway, hopefully that made sense and sorry I'm being confusing!

rwblickhan: (But yes, please post output from whatever changes you make, if it's convenient - it's a bit hard to tell how the Slack formatting will get translated to actual Slack messages.)

command/core.py
```diff
@@ -46,3 +50,23 @@ def handle_team_join(self, event_data):
...
+
+    def get_help(self):
+        """Get help messages and return a formatted string for messaging."""
+        message = {"text": "Displaying all available commands. To read about"
+                           " a specific command, use `/rocket COMMAND help`\n",
+                   "mrkdwn": "true"}
+        attachments = []
+        for cmd in self.__commands:
+            cmd_name = self.__commands[cmd].get_name()
+            cmd_text = "*[" + cmd_name.upper() + "]*\n\n" + \
```
rwblickhan: Oh, one other small thing - I'm personally not a fan of the `[ ]` around the command name, since that's often used for optional arguments and stuff like that. Is it possible to remove those?

command/core.py
ravenCondol: okay i think i get it; the description is neither the usage nor the help of each command, just like text to what it does

rwblickhan: Yep, which is basically what you already have with the "Commands for editing users" line... I just want each command to have its own description 😛

command/core.py
ravenCondol: yep, should i remove it in the /rocket code block in the picture above too?

rwblickhan: Hmm, I think it's more okay there - I guess I shouldn't have said "optional arguments", but more like "things where you're supposed to fill in what you want and/or optional arguments." So in `/rocket [command] help` it makes sense, but not so much in `[User]`. 

... these things are hard to describe 😂

rwblickhan: Oh, one other thing I just noticed - I think it would make more sense to print out the command name that you actually type, i.e. without the uppercase.

ravenCondol: oki i fixed it 
image

ravenCondol:
i realize that core is not currently aware of the team command module yet, and so will not be displayed if the user writes /rocket help

rwblickhan:
Okay, no problem - I believe what you've written will automatically pick it up as soon as we connect `TeamCommand`. Are you going to add a description to `TeamCommand` as well, though, just so we don't forget it?

ravenCondol:
okay, prob'll reduce coverage by a little bit though haha. 

also, the similarities between team and user really makes the java part of me want to have some sort of command abstract superclass, not sure if that's something really needed in python though

rwblickhan:
Funny you should say that... we've had #65 for very very long time, where we wanted to use a metaclass to ensure the specification of `Command` class... so not quite a superclass, but more like an interface, but that should give us a bit more control over how exactly each `Command` class looked. Unfortunately, other things have taken precedence... but we could definitely look into that again in the future.

command/core.py
```diff
@@ -46,3 +50,18 @@ def handle_team_join(self, event_data):
...
+                           "\n`/rocket [command] help`\n",
+                   "mrkdwn": "true"}
+        attachments = []
+        for cmd in self.__commands.values():
+            cmd_name = cmd.get_name()
+            cmd_text = "*" + cmd_name + ":* " + cmd.get_desc()
+            attachment = {"text": cmd_text, "mrkdwn_in": ["text"]}
+            attachments.append(attachment)
+        message["attachments"] = attachments
+        return json.dumps(message)
```
cheukyin699: We should be using `flask.jsonify` instead of `json.dumps` so that we can get the correct headers for the response.

tests/command/core_test.py
```diff
@@ -33,6 +35,26 @@ def test_handle_invalid_command(mock_logging, mock_usercommand):
     error_txt = "Please enter a valid command."
 
 
+@mock.patch('command.core.logging')
+def test_handle_help(mock_logging, ):
```
cheukyin699: Coma?

ravenCondol: fixed

command/core.py
ravenCondol: done !

tests/command/core_test.py
```diff
@@ -33,6 +35,31 @@ def test_handle_invalid_command(mock_logging, mock_usercommand):
...
+        expect = json.loads(
+            jsonify({"text": "Displaying all available commands. "
+                             "To read about a specific command, "
+                             "use \n`/rocket [command] help`\n",
+                     "mrkdwn": "true",
+                     "attachments": [
+                         {"text": "*user:* for dealing with users",
+                          "mrkdwn_in": ["text"]}]}).data)
+        resp = json.loads(resp.data)
+    print(resp)
```
omar2535: why u print

ravenCondol: nnnnnnnnoooooooo

PR #211, opened 2019-02-09
Fix #207 - admin promotion feature

### LINKED ISSUE #207: Add admin-level command to make other users admin
Right now I believe we have to edit a user directly in the database to make them an admin. Clearly this won't be sustainable. We should add an admin-level subcommand to `/rocket user` that changes their permission level - possibly only upwards, to prevent infighting.

cheukyin699:
# Pull Request

## Description

Add permission level promotion feature with docs to back it up

## Ticket(s)

Closes #207.

### DIFF
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -75,7 +75,7 @@ class UserCommand:
         parser_edit.add_argument("--major", type=str, action='store')
         parser_edit.add_argument("--bio", type=str, action='store')
         parser_edit.add_argument("--member", type=str, action='store')
-        parser_edit.add_argument("--perm", type=lambda x: Permissions[x],
+        parser_edit.add_argument("--permission", type=lambda x: Permissions[x],
                                  action='store', choices=list(Permissions))
 
     def get_name(self):
@@ -123,7 +123,7 @@ class UserCommand:
                 "github": args.github,
                 "major": args.major,
                 "bio": args.bio,
-                "perm": args.perm,
+                "perm": args.permission,
             }
             return self.edit_helper(user_id, param_list)
 
@@ -181,8 +181,7 @@ class UserCommand:
         if param_list["perm"] and is_admin:
             edited_user.set_permissions_level(param_list["perm"])
         elif param_list["perm"] and not is_admin:
-            msg += "\nCannot change user permission: user isn't admin;" +\
-                " this incident will be reported."
+            msg += "\nCannot change own permission: user isn't admin."
             logging.warn("User {} tried to elevate permissions level."
                          .format(user_id))
 
--- a/command/commands/user.py
+++ b/command/commands/user.py
@@ -123,7 +123,7 @@ class UserCommand:
                 "github": args.github,
                 "major": args.major,
                 "bio": args.bio,
-                "perm": args.permission,
+                "permission": args.permission,
             }
             return self.edit_helper(user_id, param_list)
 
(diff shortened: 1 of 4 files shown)


command/commands/user.py
```diff
@@ -168,6 +173,13 @@ def edit_helper(self, user_id, param_list):
             edited_user.set_major(param_list["major"])
         if param_list["bio"]:
             edited_user.set_biography(param_list["bio"])
+        if param_list["perm"] and is_admin:
+            edited_user.set_permissions_level(param_list["perm"])
+        elif param_list["perm"] and not is_admin:
+            msg += "\nCannot change user permission: user isn't admin;" +\
+                " this incident will be reported."
+            logging.warn("User {} tried to elevate permissions level."
```
rwblickhan: Might be valuable to also include which user they tried to elevate permissions on.

command/commands/user.py
```diff
@@ -74,6 +74,8 @@ def init_subparsers(self):
         parser_edit.add_argument("--major", type=str, action='store')
         parser_edit.add_argument("--bio", type=str, action='store')
         parser_edit.add_argument("--member", type=str, action='store')
+        parser_edit.add_argument("--perm", type=lambda x: Permissions[x],
```
rwblickhan: I find `--perm` a little bit non-obvious - I feel like "perm" isn't really a common shortening of "permissions."

docs/UserCommands.md
```diff
@@ -39,6 +40,12 @@ another user's Launch Pad profile by using `[--member SLACKID]` option.
 /rocket user edit --member @s_universe --name "Steven Universe"
 ```
 
+Admins can easily create other admins or team leads.
```
rwblickhan: Maybe "promote" instead of "create"?

command/commands/user.py
cheukyin699: If the user is not an admin, they aren't allowed to use `--member`, so they can only use this command on themselves. All other instances will result in a "You don't have permissions to do so." error.

rwblickhan: Er, I don't follow. When will this case even happen, then?

cheukyin699: When the user self-promotes

rwblickhan: Oh I see. Perhaps something like "user cannot change own permission" or something? I feel like this message doesn't make it obvious that the user is trying to change their own permission level.

command/commands/user.py
michaelenglo: I agree. I would just put `--permissions` instead

michaelenglo: since you are not looking at your permissions that often anyway, it doesn't hurt to write the whole thng.

docs/UserCommands.md
```diff
@@ -25,6 +25,7 @@ overwrite users that have already been entered into the database. By using the
 /rocket user edit [--name NAME] [--email EMAIL] [--pos POSITION]
                   [--github GITHUB_HANDLE] [--major MAJOR]
                   [--bio BIOGRAPHY]
+                  [--perm {member,team_lead,admin}]
```
rwblickhan: This should be updated to reflect `--permission`.

command/commands/user.py
```diff
@@ -121,6 +123,7 @@ def handle(self, command, user_id):
                 "github": args.github,
                 "major": args.major,
                 "bio": args.bio,
+                "perm": args.permission,
```
omar2535: I think this should be changed to be `permission` just so that we are consistent with all the other arguments and their associated parameter. 
```
        parser_edit.add_argument("--name", type=str, action='store')
        parser_edit.add_argument("--email", type=str, action='store')
        parser_edit.add_argument("--pos", type=str, action='store')
        parser_edit.add_argument("--github", type=str, action='store')
        parser_edit.add_argument("--major", type=str, action='store')
        parser_edit.add_argument("--bio", type=str, action='store')
        parser_edit.add_argument("--member", type=str, action='store')
        parser_edit.add_argument("--permission", type=lambda x: Permissions[x],
                                 action='store', choices=list(Permissions))
```
and
```
            param_list = {
                "member": args.member,
                "name": args.name,
                "email": args.email,
                "pos": args.pos,
                "github": args.github,
                "major": args.major,
                "bio": args.bio,
                "perm": args.permission,
            }
```

Issue #163, opened 2019-01-12
Create team member Github API interface


rwblickhan:

michaelenglo:
Functionality added:
-  list team member
-  get team member
-  add team member
-  remove team member

rwblickhan:
Closed by #163.

PR #213, opened 2019-02-10
#163 (Add Github TeamMember interface)

michaelenglo:
# Pull Request

## Description

## Testing

## Ticket(s)
#163

Affects #


interface/github.py
```diff
@@ -45,6 +45,49 @@ def org_has_member(self, username):
...
 
+# ---------------------------------------------------------------
+# --------------- methods related to team members ---------------
+# ---------------------------------------------------------------
+
+    def list_team_members(self, team_id):
+        """Return a list of users in the team of id team_id."""
+        try:
+            team = self.github.get_team(team_id)
+            # Question: Should we return github.PaginatedList.PaginatedList or just regular python's list
```
rwblickhan: On this, check with @jacqueschen1 - I believe he converts `PaginatedList` to a standard list, but I can't quite remember.

jacqueschen1: yea you can iterate through the `PaginatedList` and put everything in a Python list.

interface/github.py
```diff
@@ -45,6 +45,54 @@ def org_has_member(self, username):
...
+# --------------- methods related to team members ---------------
+# ---------------------------------------------------------------
+
+    def list_team_members(self, team_id):
+        """Return a list of users in the team of id team_id."""
+        try:
+            team = self.github.get_team(team_id)
+            # Question: Should we return
+            # github.PaginatedList.PaginatedList or just
+            # regular python's list.
```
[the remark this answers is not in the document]
michaelenglo: Actually now that I think about it, I think it seems like it should be okay to return `PaginatedList` since it has the same interface as normal python's list and you don't have to import anything to be able to use it. Let's talk about it when ur here.

tests/interface/github_test.py
```diff
@@ -16,6 +16,20 @@ def setUp(self):
...
 
+        # make mock team
+        self.mock_team = mock.MagicMock(Team.Team)
+        # right now, any arg team_id to get_team()
+        # will return self.mock_team no mater what.
+        # this is not robust since get_team should
+        # return different teams depending on the team_id
+        # being passed in. If anyone has a better
+        # idea, please let me know.
+        self.mock_github.get_team = mock.MagicMock(return_value=self.mock_team)
```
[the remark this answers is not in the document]
michaelenglo: thanks that's exactly what I need!

michaelenglo:
I have resolved all the comments but I have test coverage error. Here is a snippet of the log error from [travis](https://travis-ci.org/ubclaunchpad/rocket2/builds/494310841?utm_source=github_status&utm_medium=notification)

```
==> Detecting CI provider
    Travis Detected
==> Preparing upload
==> Processing gcov (disable by -X gcov)
    Executing gcov (find /home/travis/build/ubclaunchpad/rocket2 -not -path './bower_components/**' -not -path './node_modules/**' -not -path './vendor/**' -type f -name '*.gcno'  -exec gcov -pb  {} +)
==> Collecting reports
Error: No coverage report found
```

Can somebody help me figure out why this happened?

interface/github.py
```diff
@@ -105,6 +105,51 @@ def org_get_teams(self):
...
 
+# ---------------------------------------------------------------
+# --------------- methods related to team members ---------------
+# ---------------------------------------------------------------
+
+    def list_team_members(self, team_id):
+        """Return a list of users in the team of id team_id."""
+        try:
+            team = self.github.get_team(team_id)
+            return map(lambda x: x, team.get_members())
```
cheukyin699: This does not return a list. Instead, this returns a Map iterator (or a map object, whatever you like to call it). The difference between the two is that you cannot index a Map object. This is a Python 3 thing. Please convert it to a list:

```python
list(map(lambda x: x, team.get_members))
```

interface/github.py
```diff
@@ -105,6 +105,51 @@ def org_get_teams(self):
...
+        except GithubException as e:
+            raise GithubAPIException(e.data)
+
+    def get_team_member(self, member_username, team_id):
```
omar2535: Is there a reason why this is called `member_username` but the others are called `username`?

interface/github.py
```diff
@@ -105,6 +105,51 @@ def org_get_teams(self):
...
+                "user \"{}\" does not exist in team \"{}\""
+                .format(member_username, team_id))
+
+    def add_team_member(self, username, team_id):
+        """Add user with given username to team with id team_id."""
+        try:
+            team = self.github.get_team(team_id)
+            new_member = self.github.get_user(username)
+            team.add_membership(new_member)
+        except GithubAPIException as e:
```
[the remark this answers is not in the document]
rwblickhan: Good catch, I believe you're correct.

interface/github.py
```diff
@@ -105,6 +105,51 @@ def org_get_teams(self):
...
+            raise GithubAPIException(e.data)
+
+    def remove_team_member(self, username, team_id):
+        """Remove user with given username from team with id team_id."""
+        try:
+            team = self.github.get_team(team_id)
+            to_be_removed_member = self.github.get_user(username)
+            team.remove_membership(to_be_removed_member)
+        except GithubAPIException as e:
```
rwblickhan: Also down here - I think this should be `GithubException` as well.

interface/github.py
michaelenglo: didn't know that. Good call

interface/github.py
michaelenglo: yes, thank you. I think it's there to indicate that the username has to be an existing team member, not just any username, otherwise, it will throw an exception. But to maintain consistency, I will change this to `username`.

interface/github.py
michaelenglo: Yes, thank you!

interface/github.py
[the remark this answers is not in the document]
michaelenglo: Ah, that was a mistake. Good catch!

PR #216, opened 2019-02-11
Rwblickhan/#215 add rocket token command

### LINKED ISSUE #215: Add `/rocket token` command
Add a `/rocket token` command that generates and returns a signed [JSON Web Token](https://jwt.io/) for the user, which can then be used to authenticate with a potential HTTP API in the future.

rwblickhan:
# Pull Request

## Description

Add a `/rocket token` command that generates and presents a signed JWT for the user, for use with a potential future HTTP API.

## Testing

See picture of output in comments below.

## Ticket(s)

Closes #215

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -17,6 +17,7 @@ colorama = "*"
 flask-talisman = "*"
 flask-seasurf = "*"
 pyjwt = "*"
+flask-limiter = "*"
 
 [dev-packages]
 awscli = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -16,6 +16,7 @@ structlog = "*"
 colorama = "*"
 flask-talisman = "*"
 flask-seasurf = "*"
+pyjwt = "*"
 
 [dev-packages]
 awscli = "*"
--- a/command/commands/token.py
+++ b/command/commands/token.py
@@ -1,0 +1,59 @@
+"""Command to obtain signed authentication token."""
+import logging
+import jwt
+
+from model.permissions import Permissions
+from datetime import datetime, timedelta
+
+
+class TokenCommand:
+    """Token command model class."""
+
+    command_name = "token"
+    desc = "Generate a signed token for use with the HTTP API"
+    permission_error = "You do not have the sufficient " \
+                       "permission level for this command!"
+    lookup_error = "Requesting user not found!"
+
+    def __init__(self, db_facade):
+        """
+        Initialize TokenCommand.
+
+        :param db_facade: Database connection
+        """
+        logging.info("Initializing TokenCommand instance")
+        self.facade = db_facade
+        self.key = ";alskdjf;alh;alsdjf;aklsdjfl;"
+
+    def get_name(self):
+        """Return the command's name."""
+        return self.command_name
+
+    def get_desc(self):
+        """Return a description of the command."""
+        return self.desc
+
+    def handle(self, _command, user_id):
+        """Handle request for token."""
+        logging.debug("Handling token command")
+        try:
+            lead_user = self.facade.retrieve_user(user_id)
+            if lead_user.get_permissions_level() == Permissions.member:
+                return self.permission_error, 403
+        except LookupError:
+            return self.lookup_error, 404
+        expiry = datetime.utcnow() + timedelta(days=7)
+        payload = {
+            'nbf': datetime.utcnow(),
+            # TODO make this configurable
+            'exp': expiry,
+            'iss': 'ubclaunchpad:rocket2',
+            'iat': datetime.utcnow(),
+            'user_id': user_id
+        }
+        token = jwt.encode(payload, self.key, algorithm='HS256') \
+            .decode('utf-8')
+        return "This is your token:\n```\n{}".format(token) + \
+               "\n```\nKeep it secret! Keep it safe!\n" \
+               "It will expire at {}.".format(expiry), \
+               200
--- a/command/commands/token.py
+++ b/command/commands/token.py
@@ -14,12 +14,15 @@ class TokenCommand:
     permission_error = "You do not have the sufficient " \
                        "permission level for this command!"
     lookup_error = "Requesting user not found!"
+    success_msg = "This is your token:\n```\n{}\n```" \
+                  "\nKeep it secret! Keep it safe!\nIt will expire at {}."
 
     def __init__(self, db_facade, config):
         """
         Initialize TokenCommand.
 
         :param db_facade: Database connection
+        :param config: :class:`command.commands.TokenCommandConfig` object
         """
         logging.info("Initializing TokenCommand instance")
         self.facade = db_facade
@@ -54,10 +57,8 @@ class TokenCommand:
         }
         token = jwt.encode(payload, self.signing_key, algorithm='HS256') \
             .decode('utf-8')
-        return "This is your token:\n```\n{}".format(token) + \
-               "\n```\nKeep it secret! Keep it safe!\n" \
-               "It will expire at {}.".format(expiry), \
-               200
+        format(token)
+        return self.success_msg.format(token, expiry), 200
 
 
 class TokenCommandConfig:
--- a/command/commands/token.py
+++ b/command/commands/token.py
@@ -26,23 +26,15 @@ class TokenCommand:
         """
         logging.info("Initializing TokenCommand instance")
         self.facade = db_facade
-        self.expiry = config.get_expiry()
-        self.signing_key = config.get_signing_key()
-
-    def get_name(self):
-        """Return the command's name."""
-        return self.command_name
-
-    def get_desc(self):
-        """Return a description of the command."""
-        return self.desc
+        self.expiry = config.expiry
+        self.signing_key = config.signing_key
 
     def handle(self, _command, user_id):
         """Handle request for token."""
         logging.debug("Handling token command")
         try:
             user = self.facade.retrieve_user(user_id)
-            if user.get_permissions_level() == Permissions.member:
+            if user.permissions_level == Permissions.member:
                 return self.permission_error, 403
         except LookupError:
             return self.lookup_error, 404
@@ -53,7 +45,7 @@ class TokenCommand:
             'iss': 'ubclaunchpad:rocket2',
             'iat': datetime.utcnow(),
             'user_id': user_id,
-            'permissions': user.get_permissions_level().value
+            'permissions': user.permissions_level.value
         }
         token = jwt.encode(payload, self.signing_key, algorithm='HS256') \
             .decode('utf-8')
@@ -68,11 +60,3 @@ class TokenCommandConfig:
         """Initialize config for TokenCommand."""
         self.expiry = expiry
         self.signing_key = signing_key
-
-    def get_expiry(self):
-        """Return expiry offset, expressed as a `timedelta`."""
-        return self.expiry
-
-    def get_signing_key(self):
-        """Return string containing secret authentication signing key."""
-        return self.signing_key
--- a/command/commands/token.py
+++ b/command/commands/token.py
@@ -1,9 +1,9 @@
 """Command to obtain signed authentication token."""
-import logging
 import jwt
+import logging
 
+from datetime import datetime
 from model.permissions import Permissions
-from datetime import datetime, timedelta
 
 
 class TokenCommand:
@@ -15,7 +15,7 @@ class TokenCommand:
                        "permission level for this command!"
     lookup_error = "Requesting user not found!"
 
-    def __init__(self, db_facade):
+    def __init__(self, db_facade, config):
         """
         Initialize TokenCommand.
 
@@ -23,7 +23,8 @@ class TokenCommand:
         """
         logging.info("Initializing TokenCommand instance")
         self.facade = db_facade
-        self.key = ";alskdjf;alh;alsdjf;aklsdjfl;"
+        self.expiry = config.get_expiry()
+        self.signing_key = config.get_signing_key()
 
     def get_name(self):
         """Return the command's name."""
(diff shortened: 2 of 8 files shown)


rwblickhan:
screen shot 2019-02-10 at 8 49 46 pm

^ Output.

command/commands/token.py
```diff
@@ -0,0 +1,77 @@
...
+        self.expiry = config.get_expiry()
+        self.signing_key = config.get_signing_key()
+
+    def get_name(self):
+        """Return the command's name."""
+        return self.command_name
+
+    def get_desc(self):
+        """Return a description of the command."""
+        return self.desc
```
bobheadxi: this is more of an aside, but TIL lots of getters is a thing in python, I thought the meme was that java's getters are half of its verbosity?

command/commands/token.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def handle(self, _command, user_id):
+        """Handle request for token."""
+        logging.debug("Handling token command")
```
bobheadxi: one thing I've discovered is that logs like this are *super* unhelpful, because you end up getting a mumbo jumbo of logs with no way of associating series of events. For example, 2 users calling the same endpoint might generate logs like this:

```
DEBUG handling
DEBUG handling
INFO handling with option
INFO handling without option
INFO something
WARNING oh no
INFO success
INFO success
```

gets weird trying to tell who's who - it might be helpful to have some sort of request ID that propagates down to functions and can be included in logs as a field

command/commands/token.py
```diff
@@ -0,0 +1,77 @@
...
+            'iss': 'ubclaunchpad:rocket2',
+            'iat': datetime.utcnow(),
+            'user_id': user_id,
+            'permissions': user.get_permissions_level().value
+        }
+        token = jwt.encode(payload, self.signing_key, algorithm='HS256') \
+            .decode('utf-8')
+        return "This is your token:\n```\n{}".format(token) + \
+               "\n```\nKeep it secret! Keep it safe!\n" \
+               "It will expire at {}.".format(expiry), \
```
bobheadxi: this feels a little tricky to read 😢 

it might help to have a set of slack/markdown util functions like this:

```python
def code(str):
    return "`{}`".format(str)

def code_block(str):
    return "```\n{}\n```".format(str)

def quote(str):
    return "> {}\n".format(str)

def emph(str):
    return "*{}*".format(str)
```

etc, then this could be:

```python
return "This is your token: {}".format(slack_fmt.code_block(token))
```

bobheadxi: (since I imagine such formatting will probably be used frequently throughout rocket)

Pipfile
```diff
@@ -14,6 +14,8 @@ pygithub = "*"
 toml = "*"
 structlog = "*"
 colorama = "*"
+pyjwt = "*"
+flask-limiter = "*"
```
bobheadxi: this doesn't look like it's used in this PR yet?

rwblickhan: No, I was planning to start working on that but figured I should get this reviewed first. I don't think it's worth removing at this point though, because it definitely _will_ be used at some point.

command/commands/token.py
rwblickhan: I'm not sure that a lot of getters is actually a thing in Python - thoughts, @cheukyin699?

command/commands/token.py
rwblickhan: Definitely a good idea, but low priority I would say - I'll make a ticket.

rwblickhan: (Unless _you_ want to make a ticket. 😛)

bobheadxi: haha sure ill do it, but yeah i actually feel this is a pretty high priority, since the longer you wait the more you'll have to replace and the less likely anyone will want to do it 😋

bobheadxi: see #218

rwblickhan: Doesn't matter if they want to do it if it's their assigned task 😉 

And while it is somewhat hard to read, I don't think it's _that_ important to fix - I think it's relatively obvious what the output is supposed to be, even if the exact formatting is a bit hard to parse.

rwblickhan: Thanks for making the new ticket though!

bobheadxi: Hard to parse == easy to make mistakes 😰 I updated the example:

```py
return "This is your token: {}".format(slackfmt.code_block(token)) + \
    "Keep it secret! Keep it safe!\n" + \
    "It will expire at {}.".format(expiry), \
    200
```

I'd say removing the "noise" of various markdown annotations makes it a lot clearer

command/commands/token.py
rwblickhan: Honestly `user_id` would probably work for that - that's passed directly from Slack to us, so it's as good as a request ID (and I highly doubt somebody would try spamming commands to such a point we can't tell their requests apart).

command/commands/token.py
rwblickhan: But you've already done the work of figuring out what it does, and we never need to touch it again, right 😛

bobheadxi: Or:

```py
return """
This is your token: {}".format(slackfmt.code_block(token))
Keep it secret! Keep it safe!
It will expire at {}.".format(expiry)
    """,
    200
```

(untested, but i think that would work)

bobheadxi: 😢

command/commands/token.py
bobheadxi: that is, until we have users programmatically accessing APIs with tokens and whatnot

rwblickhan: But this is _strictly_ for the Slack command - somebody accessing programmatically won't ever arrive here.

bobheadxi: oh, i was under the assumption that whatever API you guys build my hook into the command framework to avoid rewriting functionality, like:

```sh
rocket2.ubclaunchpad.com/user/edit?member=rwblickhan
# /rocket user edit --member=rwblickhan
```

rwblickhan: Probably something like that - though I was specifically referring to `/rocket token`, which should reaaaally only be accessible via Slack 😛

bobheadxi: tricky though - ideally an API would return proper JSON structures (we return plain text in inertia and https://github.com/ubclaunchpad/inertia/pull/546 was... not fun to write...)

I might consider having these functions return objects with relevant information, like

```
{ token: "", expiry: "" }
```

then write a thin layer above the commands that handles things in one of two ways:

* throw the object straight to the caller if via API
* format object into a pretty object if via Slack

bobheadxi: > specifically referring to `/rocket token`, which should reaaaally only be accessible via Slack

oh yeah that's fair, I guess this is going on a bit more of a tangent about things in general 😋 just throwing some thoughts out here

rwblickhan: > then write a thin layer above the commands that handles things in one of two ways:
> 
>     * throw the object straight to the caller if via API
> 
>     * format object into a pretty object if via Slack

I was thinking something along these lines, yeah. That would require a bit of refactoring, but hopefully not _too_ much.

command/commands/token.py
cheukyin699: I think he's referring to the `@property` decorator, where you could make a function with that decorator and have slightly easier access to the variables within. Also see `@var.setter` and `@var.getter`.

A slight caveat is the classes that implement this must be children of the mighty `object` class. But nothing a bit of refactoring can't do.

rwblickhan: Sounds like another ticket. 😛

command/commands/token.py
michaelenglo: Fyi, I did some stack overflow on this topic. We can write it like this:
```
'''line {0}
line {1}
line {2}'''.format(1,2,3)
```
and this will print `'line 1\nline 2\nline 3'`.

Might help make it less ugly(?) 😕

cheukyin699: If we use that, we would destroy the indentation. In my opinion, it would make it uglier.

rwblickhan: I tend to agree... it would work, but @bobheadxi's utility solution is probably "cleaner."

Actually, now that I think about it, there's probably going to be other Launch Pad Python projects that would benefit from those functions. Perhaps we should break it out into a separate repo...?

bobheadxi: you mean a package like `slackfmt`? 😍

rwblickhan: Slack and possibly general Markdown formatting as well, yeah. That should be relatively easy to build/test.

cheukyin699: > *Grumble grumble* you teams and your multiple repositories/projects

> Something something something lisp

It sounds like a small side-project more though. Might work if we add more features.

bobheadxi: `mdfmt` 🤔

rwblickhan: Maybe `mdformatter` instead? 😛

bobheadxi: grumble grumble library names should short for easier imports grumble grumble

rwblickhan: Aren't you using an IDE with autocomplete? 😛 

Besides, `fmt` makes me think of auto-formatters like `gofmt` and `rustfmt`, which this very much is not.

bobheadxi: yeah but go's format-printing standard library is called `fmt`

```go
import "fmt"

func main() {
    var truth = "mdfmt is best"
    fmt.Printf("%s", truth)
}
```

rwblickhan: Hmm... I imagine we'd put both Markdown and Slack formatting into the same library though... so perhaps `lpfmt`? 😛

bobheadxi: Slack formatting is only a subset of Markdown, right? theres nothing in slack formatting that's not in MD

the only thing I'm aware of is italics are `_blah_` instead of `*blah*`, and bold is `*blah*` instead of `**blah**`

rwblickhan: 😱 Seems like a pretty big difference!

bobheadxi: potentially cool: a flag that toggles between MD and Slack mode, so if you try to `as_table()` with the Slack mode on, isntead of a markdown table you'll get a markdown-table-in-code-block

rwblickhan: See, I was thinking it would be easier to just have separate modules (in the same package) for Markdown and Slack formatting...

bobheadxi: I wrote a small slack module a long time ago - the entirety of its formatting can be encompassed in <50 lines though:

https://gist.github.com/bobheadxi/92737e761d60072ca905dfa3dd096cc3

rwblickhan: Yeah, this doesn’t have to be complicated, I just don’t think we should have it stuck in the Rocket 2 repository rather than, say, on PyPI 😛

bobheadxi: oh i was under the impression you were thinking of spinning it off into a new repo 😋 well in that case keep it simple 😅

rwblickhan: Yeah, I would spin it off into a new repo, just to keep it separate/distinct from all the Rocket 2 stuff.

rwblickhan: I don’t think it has to be particularly complicated, even if it’s in a new repo.

command/core.py
```diff
@@ -9,13 +10,14 @@
...
 
-    def __init__(self, db_facade, bot, gh_interface):
+    def __init__(self, db_facade, bot, gh_interface, token_config):
         """Initialize the dictionary of command handlers."""
         self.__commands = {}
         self.__facade = db_facade
         self.__bot = bot
         self.__github = gh_interface
         self.__commands["user"] = UserCommand(self.__facade, self.__github)
+        self.__commands["token"] = TokenCommand(self.__facade, token_config)
```
jacqueschen1: Is it time to finally start looking into meta-classes? :p

rwblickhan: You don't have a task anymore, right? You could definitely start looking into that now, if you'd like.

jacqueschen1: I thought I was supposed to do team commands as well? But yea

tests/command/commands/token_test.py
```diff
@@ -0,0 +1,83 @@
...
+                             "secret",
+                             algorithms='HS256')
+        assert decoded['user_id'] == user.get_slack_id()
+        assert decoded['permissions'] == Permissions.team_lead.value
+        assert ret_code == 200
+
+    def test_handle_admin(self):
+        """Test handle() when given a user with admin permissions."""
+        user = User("U12345")
+        user.set_permissions_level(Permissions.admin)
```
cheukyin699: Didn't you make an admin user in `setUp()`?

rwblickhan: Oops, my mistake, it should be done here. I'll fix that.

command/commands/token.py
```diff
@@ -0,0 +1,77 @@
+"""Command to obtain signed authentication token."""
+import jwt
+import logging
+
+from datetime import datetime
+from model.permissions import Permissions
+
+
+class TokenCommand:
+    """Token command model class."""
```
ravenCondol: I noticed that we don't initialize parsers here like in the other commands which confuses me ? Or is this something that is to be added later

rwblickhan: We don't need parsers, there's no subcommands/options.

cheukyin699:
oh that's what that button does

command/commands/token.py
```diff
@@ -0,0 +1,77 @@
...
+    desc = "Generate a signed token for use with the HTTP API"
+    permission_error = "You do not have the sufficient " \
+                       "permission level for this command!"
+    lookup_error = "Requesting user not found!"
+
+    def __init__(self, db_facade, config):
+        """
+        Initialize TokenCommand.
+
+        :param db_facade: Database connection
```
cheukyin699: Add config object to parameters

command/commands/token.py
cheukyin699: This success message should be a static constant:

```python
success_msg = "This is your token..."
# ...
success_msg.format(token, expiry, ...)
```

factory/__init__.py
```diff
@@ -17,17 +22,27 @@ def make_core(config, gh=None):
...
             config['slack']['creds_path'])['api_token']
         github_api_token = toml.load(
             config['github']['creds_path'])['api_token']
         github_organization = config['github']['organization']
         gh = GithubInterface(Github(github_api_token), github_organization)
+        if os.path.isfile(config['auth']['signing_key_path']):
+            signing_key = open(config['auth']['signing_key_path']).read()
+        else:
+            signing_key = ''.join(
+                random.choice(string.ascii_lowercase) for _ in range(24))
```
cheukyin699: Abstract into another function?

command/commands/token.py
rwblickhan: Whoops, will fix.

command/commands/token.py
rwblickhan: Will fix.

factory/__init__.py
rwblickhan: Sure.

rwblickhan:
Issue for documentation in #244.

PR #223, opened 2019-02-16
Raven c/#202 submodule help

### LINKED ISSUE #202: Refactor individual command help text delivery

ravenCondol:
# Pull Request

## Description

Individual help commands, e.g. `/rocket user help` now print the usage of every subcommand using argparse.

## Ticket(s)

Closes #202

### DIFF
--- a/command/commands/team.py
+++ b/command/commands/team.py
@@ -26,70 +26,70 @@ class TeamCommand:
         """Parser for list command."""
         parser_list = subparsers.add_parser("list")
         parser_list.set_defaults(which="list",
-                                 help="outputs the Github team names "
-                                      "and displays names of all teams")
+                                 help="Outputs the Github team names "
+                                      "and displays names of all teams.")
 
         """Parser for view command."""
         parser_view = subparsers.add_parser("view")
         parser_view.set_defaults(which="view",
-                                 help="view information and members of "
-                                      "a team")
+                                 help="View information and members of "
+                                      "a team.")
         parser_view.add_argument("team_name", type=str, action='store')
 
         """Parser for delete command."""
         parser_delete = subparsers.add_parser("delete")
         parser_delete.set_defaults(which="delete",
-                                   help="permanently delete specified "
-                                        "team (admin only)")
+                                   help="(Admin only) Permanently delete "
+                                        "specified team.")
         parser_delete.add_argument("team_name", type=str, action='store')
 
         """Parser for create command."""
         parser_create = subparsers.add_parser("create")
         parser_create.set_defaults(which="create",
-                                   help="create a new team with the "
+                                   help="(Admin only)"
+                                        "Create a new team with the "
                                         "Github team name and provided "
                                         "optional parameters. NOTE: "
-                                        "you will be added to this team "
-                                        "(admin only)")
+                                        "you will be added to this team.")
         parser_create.add_argument("team_name", type=str, action='store',
-                                   help="Github name of your team (required)")
+                                   help="Github name of your team (required).")
         parser_create.add_argument("--name", type=str, action='store',
-                                   help="display name of your team")
+                                   help="Display name of your team.")
         parser_create.add_argument("--platform", type=str, action='store',
-                                   help="the team's main platform")
+                                   help="The team's main platform.")
         parser_create.add_argument('--channel', action='store_true',
-                                   help="add all members of this channel "
-                                        "to the created team")
+                                   help="Add all members of this channel "
+                                        "to the created team.")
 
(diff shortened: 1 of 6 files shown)


ravenCondol:
Some pictures of what /rocket user help would output:
image

cheukyin699:
Just to be clear, this is output from `/rocket user help` and not `/rocket team help`?

ravenCondol:
> 
> 
> Just to be clear, this is output from `/rocket user help` and not `/rocket team help`?

yep, do you want to see the team one too? its basically the same though haha

command/commands/team.py
```diff
@@ -8,97 +8,89 @@ class TeamCommand:
...
 
     def init_subparsers(self):
         """Initialize subparsers for team command."""
         subparsers = self.parser.add_subparsers(dest="which")
 
         """Parser for list command."""
         parser_list = subparsers.add_parser("list")
-        parser_list.set_defaults(which="list")
+        parser_list.set_defaults(which="list",
+                                 help="outputs the Github team names "
```
rwblickhan: I would prefer if the help strings were capitalized and had punctuation.

rwblickhan: Not a change to be made, but I wonder if there's a way to get these help strings programmatically - perhaps there's a way to read the docstrings directly?

command/commands/team.py
```diff
@@ -8,97 +8,89 @@ class TeamCommand:
...
 
         """Parser for create command."""
         parser_create = subparsers.add_parser("create")
-        parser_create.set_defaults(which="create")
-        parser_create.add_argument("team_name", type=str, action='store')
-        parser_create.add_argument("--name", type=str, action='store')
-        parser_create.add_argument("--platform", type=str, action='store')
-        parser_create.add_argument('--channel', action='store_true')
+        parser_create.set_defaults(which="create",
+                                   help="create a new team with the "
```
rwblickhan: I think `(admin only)` should be moved to the front of the help string, so you can easily see what the permissions are when scanning the list of commands.

command/commands/team.py
```diff
@@ -110,7 +102,11 @@ def get_desc(self):
 
     def get_help(self):
         """Return command options for team events."""
-        return self.help
+        res = "\n*" + self.command_name + " commands:*```"
```
rwblickhan: Would it be possible to capitalize the command names? I think that would look a bit more professional.

rwblickhan:
Looks good! Just a few small formatting changes that I think would add that 110% extra polish 😛

command/commands/team.py
rwblickhan: Also, `admin` should be capitalized.

command/commands/team.py
ravenCondol: there is a way yep, actually i think that's what my get_help is doing, just iterating over each command and appending their individual help strings

command/commands/team.py
ravenCondol: should the first letter after the ending bracket be capitalized?

command/commands/team.py
rwblickhan: Sorry, I mean, it looks like we have to set this `help` parameter by hand. I'm wondering if there's a way to set it in an automated fashion, perhaps by reading the docstrings we write for each handler directly. This would probably be quite a big task though, so I wouldn't bother with it for now.

command/commands/team.py
rwblickhan: I think so. I'd prefer something like:
```
create: (Admin only) Create a new team with...
```
But I could see arguments for other capitalizations as well 😛

command/commands/team.py
ravenCondol: ok

command/commands/team.py
ravenCondol: yep

command/commands/team.py
ravenCondol: nice i guessed correctly. that is what it should look like now

command/commands/team.py
ravenCondol: o ok i see

rwblickhan:
Nice! Looks good to me now. 👍

ravenCondol:
nani the heck are these conflicts?!? hol' up lemme fix em

PR #227, opened 2019-02-18
#142 slack interface functions

### LINKED ISSUE #142: Implement team commands
Here is the list of stubs to finish off:

- [x] team list
- [x] team view
- [x] team delete
- [x] team create
- [x] team add
- [x] team remove
- [x] team edit
- [x] team lead

Fill this in when you finish, and when they all are finished, close this.

jacqueschen1:
# Pull Request

## Description

Added get_channel_users function to slack interface class for future use.

## Testing

## Ticket(s)

Closes #142 

### DIFF
--- a/command/commands/team.py
+++ b/command/commands/team.py
@@ -44,7 +44,7 @@ class TeamCommand:
     desc = "for dealing with " + command_name + "s"
 
     def __init__(self, sc):
-        """Initialize team command parser."""
+        """Initialize team command parser with given Slack Client Interface."""
         logging.info("Initializing TeamCommand instance")
         self.sc = sc
         self.desc = ""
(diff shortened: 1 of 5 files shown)


command/commands/team.py
```diff
@@ -43,9 +43,10 @@ class TeamCommand:
            "    * permanently delete the specified team\n"
     desc = "for dealing with " + command_name + "s"
 
-    def __init__(self):
+    def __init__(self, sc):
         """Initialize team command parser."""
```
cheukyin699: Add docs about parameter `sc`.

command/commands/team.py
```diff
@@ -43,9 +43,10 @@ class TeamCommand:
            "    * permanently delete the specified team\n"
     desc = "for dealing with " + command_name + "s"
 
-    def __init__(self):
-        """Initialize team command parser."""
+    def __init__(self, sc):
+        """Initialize team command parser with given Slack Client Interface."""
```
rwblickhan: For future reference, we may want to move parameter documentation to a `:param:` field in the docstring - I think that's missing in many places though, so I wouldn't change it now.

Issue #201, opened 2019-02-08
Kudos system


bobheadxi:
This definitely requires a bit more thought:

* we need incentive to give kudos
* we need a good way to show kudos (leaderboards?)

Bonus points for:

* automatic kudos per PR or Issue created
    * more bonus point for a smart scaling system that reduces kudos given for tech leads and for consecutive PRs and issues
* NFC tags + mobile app, so people can tap someone with their phone and give them Rocket Kudos™️ 

- [x] Make command docs
- [x] Add into User Models
- [x] Add command

rwblickhan:
What do you mean by incentive? Isn’t that the presidents job? 😛

I think as an MVP for this feature we just need to add a kudos/karma field to a user and a command to add/remove kudos/karma (probably with some limits?).

bobheadxi:
By incentive I was thinking something automated, like... adding a `kudos-giver` flair, or a bonus multiplier on kudos you receive 🤔 

I don't think we should allow kudos removal - positive reinforcement!

rwblickhan:
Yeah I agree - the only reason my karmabot included karma removal was because it was a pretty direct clone of an existing bot that I used on... Hipchat....

Sure, we can have an automated feature, but that still sounds like something for our daring PM Robert to decide 😛

bobheadxi:
paging @bfbachmann @frostyshadows for thoughts 😋

bfbachmann:
Yea this sounds okay. The NFC tags thing sounds a bit weird, though.

cheukyin699:
Here's how other people do karma:
![image](https://user-images.githubusercontent.com/6824907/53290688-93574600-375c-11e9-83e5-0f724152ba24.png)

Just wanted to put this somewhere.

cheukyin699:
Karma done

PR #228, opened 2019-02-18
Karma doc

omar2535:
# Pull Request

## Description
Adding karma documentation. Please review & give me your thoughts.  

## Testing
N/A

## Ticket(s)
Affects #201

### DIFF
--- a/docs/KarmaCommands.md
+++ b/docs/KarmaCommands.md
@@ -9,13 +9,7 @@ Command to giveth or taketh away a user's karma
 #### Add 1 karma to user
 
 ```sh
-/rocket @user ++
-```
-
-#### Remove 1 karma from user
-
-```sh
-/rocket @user --
+/rocket karma @user
 ```
 
 ### For admin only
@@ -35,25 +29,24 @@ Command to giveth or taketh away a user's karma
 #### Reset all user karma
 
 ```sh
-/rocket karma reset all
+/rocket karma reset --all
 ```
 
-#### Examples
+### Examples
 
 ```sh
 # normal user
-/rocket @coolkid1 ++ #adds 1 karma to coolkid1
-/rocket @coolkid1 -- #removes 1 karma from coolkid1
+/rocket karma @coolkid1 #adds 1 karma to coolkid1
 
 # admin only
 /rocket karma set @coolkid1 5 #sets coolkid's karma to 5
 /rocket karma reset @coolkid1 #resets coolkid1's karma to 1
-/rocket karma reset all #resets all users karma to 1
+/rocket karma reset --all #resets all users karma to 1
 ```
 
 #### Help
 
-##### Display options for the karma commands
+##### Display options for karma commands
 
 ```sh
 /rocket karma help
--- a/docs/KarmaCommands.md
+++ b/docs/KarmaCommands.md
@@ -1,48 +1,60 @@
-# Karma Command Reference 
+# Karma Command Reference
 
 Command to giveth or taketh away a user's karma
 
 ## Options
-#### For normal users:
-##### Add 1 karma to user
-```
+
+### For normal users
+
+#### Add 1 karma to user
+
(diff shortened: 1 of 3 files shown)


docs/KarmaCommands.md
```diff
@@ -0,0 +1,60 @@
...
+Command to giveth or taketh away a user's karma
+
+## Options
+
+### For normal users
+
+#### Add 1 karma to user
+
+```sh
+/rocket @user ++
```
cheukyin699: Would this just be the `elif` statement on `command/core.py:41`, to check if the user exists or not?

cheukyin699:
Please add the karma field to the database docs in `docs/Database.md`.

docs/KarmaCommands.md
omar2535: Kinda. I was thinking of something like doing a regex match to see if it's a user id passed in then going to the karma command handlers to check if the user exists or not.

docs/KarmaCommands.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+#### Remove 1 karma from user
+
+```sh
+/rocket @user --
```
rwblickhan: Pinging @bobheadxi - he mentioned something at some point about only having "add karma," not "remove karma," to encourage people to be nice. If we only have added karma, then we'd be able to simplify this to, say, `/rocket karma @user`, which I think would be a bit more consistent.

docs/KarmaCommands.md
rwblickhan: Personally I don't really like how this would be inconsistent with other commands, especially since we'd _also_ have a `/rocket karma` command still. See my other comment though - if we take away the ability to remove karma, we'd be able to have `/rocket karma @user`, which isn't much more to type than `/rocket @user ++`.

docs/KarmaCommands.md
```diff
@@ -0,0 +1,60 @@
...
+#### Reset user karma
+
+```sh
+/rocket karma reset @user
+```
+
+#### Reset all user karma
+
+```sh
+/rocket karma reset all
```
rwblickhan: `all` should probably be a flag, like `--all`.

docs/KarmaCommands.md
bobheadxi: yeah i dont think we should be allowed to remove karma - it seems like it'll only cause problems, if people take karma seriously (hopefully not overly so). Like @rwblickhan says, it'll simplify the command too

I noticed there's a `set`, I guess for situations where things are bugged and we really need to reduce someone's karma for some reason

omar2535: Ok, I agree with removing the ability to remove karma, though changing the command to add karma might not be as intuitive.

rwblickhan: I would really like to avoid `/rocket @user ++` if possible though - I don't think the gains from intuitiveness would be worth the loss of consistency. Is there any other way we could handle this command?

bobheadxi: I really like `/rocket @user ++` since the simplicity might encourage its use more, and perhaps we can add more user-user interactions via `/rocket @user` commands. I'm thinking:

```
/rocket @user are-you-coming-to-todays-meeting
/rocket @user schedule
/rocket @user ping
```

sorta reminds me of [`inertia ${remote_name} ...`](https://inertia.ubclaunchpad.com/cli/inertia_$%7Bremote_name%7D.html) 😉 

(un-resolving this conversation because i think it warrants more discussion)

rwblickhan: Okay, I agree with this - `/rocket @user ++` will work. I think in terms of architecture we'll want to create an `InteractionCommand` (or similar) and when we parse `@user` we can pass the target user to the `InteractionCommand`. This would enable us to extend it with more "direct interaction" features in the future (as @bobheadxi mentioned above).

PR #229, opened 2019-02-18
Remove getters and setters from models

### LINKED ISSUE #221: Remove getters and setters
We use getters and setters everywhere. However, apparently this is not considered best practices in Python. Rather, we should use the simple attributes, and if we do need to change the implementation of getting/setting in the future, we can use [the `@property` decorator](https://docs.python.org/3.7/library/functions.html#property) instead. Removing the getters and setters would save us quite a bit of code.

cheukyin699:
# Pull Request

## Description

Remove the getter and setter methods from models, replacing them with a more Pythonic way of doing things: `@property` decorators!

## Ticket(s)

Closes #221

### DIFF
--- a/model/project.py
+++ b/model/project.py
@@ -19,18 +19,18 @@ class Project:
         hfunc = hashlib.sha1()
         hfunc.update(bytes(github_urls[0], 'utf-8'))
         hfunc.update(bytes(str(time.time()), 'utf-8'))
-        self.__project_id = hfunc.hexdigest()
-        self.__github_team_id = github_team_id
-        self.__github_urls = github_urls
+        self.project_id = hfunc.hexdigest()
+        self.github_team_id = github_team_id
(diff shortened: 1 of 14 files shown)


model/project.py
```diff
@@ -111,90 +111,112 @@ def __str__(self):
         """Return all fields of this project, JSON format."""
         return str(self.__dict__)
 
-    def get_project_id(self):
+    @property
```
rwblickhan: We can probably also remove most of these `@property` declarations and expose the fields directly, and only add `@property` back if we actually have to control what happens when that property is retrieved, right? That would save us quite a bit of code.

cheukyin699: Yes. That would also work. Will do.

rwblickhan:
Looks great, thanks for doing this cleanup!

PR #231, opened 2019-02-23
Refactor DynamoDB CRUD to more generic

### LINKED ISSUE #210: Refactor DynamoDB functions
DynamoDB function structures are very similar - the things different are the types of objects that they run on, the primary keys, and the table names. Everything else is literally the same.

Refactor this file so that we only have one of each function. The `store(obj)` function, `retrieve(table_name, primary_key)`, `query(table_name, parameters)`, and `delete(table_name, primary_key)`. This should cut down the number of lines in the file 3-fold.

Would like to know if we should change anything in `db/facade.py`. Opinions, @rwblickhan ?

cheukyin699:
# Pull Request

## Description

Refactor DynamoDB to remove overly specific functions and change everything else to match up with the changes. The interface for `db/facade.py` was also changed (for the better).

If I forgot to change anything, please let me know.

## Ticket(s)

Closes #210.


cheukyin699:
Blocked by #229. There could be more things that are changed that I didn't know.

db/dynamodb.py
```diff
@@ -48,6 +50,22 @@ def __init__(self, config):
         self.projects_table = config['aws']['projects_table']
         testing = config['testing']
 
+        self.TABLE_DATA[User] = self.users_table
+        self.TABLE_DATA[Team] = self.teams_table
+        self.TABLE_DATA[Project] = self.projects_table
+        self.TABLE_DATA[self.users_table] = {
+            'key': 'slack_id',
```
rwblickhan: Maybe `'key'` should be refactored into a static const string?

rwblickhan: Same for `'set_attrs'`.

PR #233, opened 2019-02-23
scaffold CI with Azure Pipelines

bobheadxi:
i put rocket in its own project https://dev.azure.com/ubclaunchpad/rocket


bobheadxi:
@rwblickhan I think I'll hand it off to you now 😅 

* `pipenv run codecov` doesn't seem to exit with 1? I ran `pipenv run codecov`, it failed `echo $?` is `0`
* codecov needs token set in the azure pipeline

Issue #224, opened 2019-02-16
Add Flask security middleware


rwblickhan:
**Please give a one-sentence summary of the feature you would like to see.**
We should add some Flask middleware that helps with security.

**Please give as many details as possible about the requirements for the feature.**
In particular, I think [Talisman](https://github.com/GoogleCloudPlatform/flask-talisman) and [SeaSurf](https://flask-seasurf.readthedocs.io/en/latest/) would be good to add.

rwblickhan:
Closed by #243.

cheukyin699:
Reopened because the last time we did this, we thought that it worked fine and it didn't. The issue was with the Flask CSRF library. The error below is what we got:

```
{Time: 2019-03-08 20:32:11, Level: [WARNING], module: flask_seasurf, function: 
_before_request():249, message: Forbidden (Referer checking failed: no referer.):
 /slack/events}
```

We don't know what's wrong so we just removed the library in a hurry.

Another thing to note is that we must set `app.secret_key` to be a random string for the session to be secure or something (probably used as a CSRF token?). To the future person on the task please do so.

And remember to test the things!

PR #243, opened 2019-02-26
installed talisman and seasurf

rmcreyes:
# Pull Request

## Description

Installed Talisman and SeaSurf and instantiated them in `server.py` so their default configurations take effect. Open to recommendations for additional configurations to set up.

## Testing

## Ticket(s)

Affects #224

### DIFF
--- a/server/server.py
+++ b/server/server.py
@@ -47,7 +47,9 @@ dictConfig({
 
 try:
     app = Flask(__name__)
+    # HTTP security header middleware for Flask
     talisman = Talisman(app)
+    # anti-CSRF middleware for Flask
     csrf = SeaSurf(app)
     config = toml.load('config.toml')
     core = make_core(config)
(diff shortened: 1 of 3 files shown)


rwblickhan:
Looks good to me. I mentioned this to you directly, but just so it's recorded somewhere, I think it's probably fine to keep the defaults enabled for both of these. However I'd be open to hearing from e.g. @bobheadxi  / @bfbachmann on this. For reference:
* [Talisman](https://github.com/GoogleCloudPlatform/flask-talisman) is HTTP security header middleware for Flask.
* [SeaSurf](https://flask-seasurf.readthedocs.io/en/latest/) is anti-CSRF middleware for Flask.

server/server.py
```diff
@@ -45,6 +47,8 @@
 
 try:
     app = Flask(__name__)
+    talisman = Talisman(app)
+    csrf = SeaSurf(app)
```
bobheadxi: some brief inline docs here might be nice because neither of these library names are self explanatory 😅

rmcreyes: aha ok i'll add that 😅

PR #245, opened 2019-03-01
Webhook events send a message to designated bot channel

### LINKED ISSUE #235: Create Slack bot channel
On startup, create a (configurable) channel where Rocket will send general information messages, if this channel doesn't exist.

### LINKED ISSUE #236: Add Slack bot messages to team webhooks
Have the Slack bot message the Slack bot channel when events occur in the team Github webhooks.

Blocked by #235.

ravenCondol:
Adds new field in config.toml for the name of the channel to send 'updates' to. 

# Pull Request

## Ticket(s)

Closes #235 
Closes #236


interface/slack.py
```diff
@@ -45,14 +44,34 @@ def get_channel_users(self, channel_id):
...
+
+    def create_channel(self, channel_name):
+        """
+        Create a channel with the given name.
+
+        :return name of newly created channel
+        """
+        logging.debug("Attempting to create channel with name {}".
+                      format(channel_name))
+        response = self.sc.api_call(
```
rwblickhan: What happens if this function is called and the channel already exists?

ravenCondol: Response will return with a 'false' in the "ok" field, and a 'name_taken' in the "error" field, I'm pretty sure

ravenCondol: o wait what do you mean, the entire function or the api_call?

rwblickhan: Well, both. If I'm understanding correctly, if the name is already taken, the API call will fail and so this will throw an error?

ravenCondol: Yep

rwblickhan: Ah, okay. In that case I think this function should handle that case and just return silently - especially since it looks like you don't catch the exception that will be thrown, so I think that would crash the whole server (unless I'm missing something?).

rwblickhan: Well, return "silently" - it should probably log a warning that the channel already exists.

ravenCondol: Oh uh yeah if the creation of the bot channel fails, the server dies. How should I handle it? Reattempt to create or just live without one

rwblickhan: Er well that is one thing to fix, but specifically what I mean here is that if a channel of the same name already exists, this function will throw an exception, which is almost definitely not the behaviour we want - the only time the channel _won't_ already exist is on the very very first launch. So I think in this case the function should notice that the channel already exists, log a warning, and just continue silently.

As to your question (i.e. if it can't create the channel for some _other_ reason), I think both of those would be appropriate - I'm not sure that retrying is that useful, though, because usually it'll be a configuration error of some kind, which won't go away if you just keep retrying 😛 So maybe just log an error and continue without, though keep in mind that'll also make any calls to post to that channel fail as well.

ravenCondol: Ohhhhh right I get it, the not_taken error should be a good thing not a bad thing

interface/slack.py
```diff
@@ -45,14 +44,33 @@ def get_channel_users(self, channel_id):
...
+            "channels.create",
+            name=channel_name,
+            validate=True
+        )
+        if 'ok' not in response or response['ok'] is False:
+            logging.error("Channel creation "
+                          "with name {} failed with error: {}".
+                          format(channel_name, response['error']))
```
rwblickhan: So my thought here is that, if this fails for a reason other than `name_taken`, it would still be a good idea to raise an exception and let the caller handle it (which is the behaviour of all the other calls) - though in this case maybe the caller just catches and ignores the exception. So I think it would be best to add a new case to the `if` statement that checks for `name_taken` and in that case logs a warning (my personal preference would be to send to `logging.warn` instead of `logging.error`, but logging levels don't really matter that much 😛) and returns, and keep this case the same as it was before (i.e. raising an exception).

ravenCondol: Hmmm I'm a little bit confused, I will ask tomorrow ahaha

tests/interface/slack_test.py
```diff
@@ -71,35 +70,42 @@ def test_send_to_channel_failure(self):
...
             assert e.error == "Error"
         finally:
             self.mock_sc.api_call.assert_called_with(
                 'conversation.members',
                 channel="C1234441"
             )
+
+    def test_create_same_channel_twice(self):
+        """Test create_channel() twice, with the second call throwing up."""
+        name: str = "#rocket2"
```
cheukyin699: Wow, types.

ravenCondol: I think pycharm added that by itself...

server/server.py
```diff
@@ -111,3 +117,14 @@ def handle_team_join(event):
     """Handle instances when user joins the Launchpad slack workspace."""
     logging.info("Handled 'team_join' event")
     core.handle_team_join(event)
+
+
+def send_webhook_notif(message):
```
rmcreyes: I was thinking maybe this method could be put in the `Bot` class seeming as it would be an appropriate layer of abstraction. Not super critical I guess but just a thought.

server/server.py
```diff
@@ -53,6 +54,9 @@
     csrf = SeaSurf(app)
     config = toml.load('config.toml')
     core = make_core(config)
+    slack_bot_channel = config['slack']['bot_channel']
```
rmcreyes: Can this be moved to `create_core()`?

server/server.py
ravenCondol: y a a , this'd mean I don't use the private __bot variable too

server/server.py
rwblickhan: I agree - reading this should be in `make_core()` with all the other ones. If you do need the `bot` out here, it's better to access it via `core.bot` (or whatever you would need there to make that work).

server/server.py
rwblickhan: Agree here too - that would also resolve the other comment.

server/server.py
ravenCondol: yes this can be done

server/server.py
ravenCondol: yep i can move `send_webhook_notif` into the `Bot` class so i dont have to mess with 'private' variables, sure. however, since team webhook notifs only have direct access to the core, i need to have a weird middle-man function in core to call the newly moved `Bot.send_webhook_notif`. is there a better way to do it?

PR #250, opened 2019-03-06
Set up Github Apps integration

### LINKED ISSUE #249: Set up Github App integration
To use the Github API, we will want to set up integration with the Github Apps API (to obtain an API token).

rwblickhan:
# Pull Request

## Description

So it turns out that we probably want to register as a Github App, which in turn requires a bit of extra logic to authenticate and then request access tokens as needed, instead of getting once permanent access token. This adds the necessary logic - in particular it makes use of a different part of the Github API, which I've put behind a new `GithubAppInterface`.

A few particular notes:
* I removed the "hack" in `server.py` to log all exceptions. I believe at some point we configured the logging "correctly" so that this is no longer necessary - or, at least, it appears to be working when I run it without that hack now.
* I cleaned up creating `GithubInterface` and `GithubAppInterface` so that it's done with a factory now.
* I moved `GithubAPIException` to a new `exceptions` sub-module, so that both `GithubInterface` and `GithubAppInterface` can access it.
* I added a few dependencies: `requests` (which is the de facto standard for HTTP requests in Python), `pem` (which is just useful for reading the PEM-format secret key file, but not really necessary), and `cryptography` (which is necessary to have the `RS256` cryptography scheme in `pyjwt`, which is required by the Github Apps API).
* I had to switch the Dockerfile to the `python:3.7` image from `python:3.7-alpine`. Unfortunately this means the final image is quite a bit larger, but it was necessary because `cryptography` is apparently unavailable for the Alpine image.

## Testing

This should Just Work™️ if you test it, but it does require that you swap out the `credentials/github.toml` for `github_signing_key.pem`, an RSA private key in PEM file format. This can be obtained from a Github Apps registration - if you want to use the one registered to `rocket2-testorg`, ask me for it (anybody with admin access to that org should also have access to it).

## Ticket(s)

Closes #249

### DIFF
--- a/interface/github.py
+++ b/interface/github.py
@@ -1,58 +1,69 @@
 """Utility classes for interacting with Github API via PyGithub."""
-from github import GithubObject, GithubException
+from github import Github, GithubObject, GithubException
+from interface.github_app import GithubAppInterface, \
+    DefaultGithubAppAuthFactory
+
+
+def handle_github_error(func):
+    """Github error handler that updates Github App API token if necessary."""
+    def wrapper(self, *arg, **kwargs):
+        try:
+            return func(self, *arg, **kwargs)
+        except GithubException as e:
+            if e.status == 401:
+                self.github = self.github_factory.create()
+                try:
+                    return func(self, *arg, **kwargs)
+                except GithubException as e:
+                    raise GithubAPIException(e.data)
+            raise GithubAPIException(e.data)
+
+    return wrapper
 
 
 class GithubInterface:
     """Utility class for interacting with Github API."""
 
-    def __init__(self, github, org):
+    def __init__(self, github_factory, org):
         """Initialize bot by creating Github object and get organization."""
+        self.github_factory = github_factory
+        self.github = github_factory.create()
         try:
-            self.github = github
             self.org = self.github.get_organization(org)
         except GithubException as e:
             raise GithubAPIException(e.data)
 
+    @handle_github_error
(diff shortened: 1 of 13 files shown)


cheukyin699:
*Extremely verbose PR - I like it!

interface/github.py
```diff
@@ -1,80 +1,90 @@
...
+    @handle_github_error
     def org_get_team(self, id):
         """Given Github team ID, return team from organization."""
-        try:
-            team = self.org.get_team(id)
-            return team
-        except GithubException as e:
-            raise GithubAPIException(e.data)
+        team = self.org.get_team(id)
+        return team
```
frostyshadows: could return `self.org.get_team(id)` directly

interface/github.py
```diff
@@ -1,80 +1,90 @@
...
+        try:
+            return func(self, *arg, **kwargs)
+        except GithubException as e:
+            if e.status == 401:
+                self.github = self.github_factory.create()
+                try:
+                    return func(self, *arg, **kwargs)
+                except GithubException as e:
+                    raise GithubAPIException(e.data)
+            raise GithubAPIException(e.data)
```
frostyshadows: this works but would be clearer if you added an `else` statement

rwblickhan: Fixed.

interface/github.py
rwblickhan: Fixed.

PR #251, opened 2019-03-09
Replaced time.time with UUID and passes tests

### LINKED ISSUE #240: Project equality test does not pass because time.time goes too slowly
**Describe the bug**

The test `test_project_equalities` does not pass for some computers because their implementation of `time.time()` is too slow.

**To Reproduce**

1. Run the test `tests/model/project_test.py`; it may or may not occur depending on your system (the only 2 reports of this are from Windows machines).

**Expected behavior**

The tests are supposed to pass.

**Screenshots**

image

**Additional context**

Windows. Windows. Windows. Windows.

michaelfu1029:
# Pull Request

## Description
Replaced time.time with uuid.uuid1() and passes tests

## Testing

No additional changes 

## Ticket(s)

closes #240


model/project.py
```diff
@@ -18,7 +20,7 @@ def __init__(self, github_team_id, github_urls):
         """
         hfunc = hashlib.sha1()
         hfunc.update(bytes(github_urls[0], 'utf-8'))
-        hfunc.update(bytes(str(time.time()), 'utf-8'))
+        hfunc.update(bytes(str(uuid.uuid1()), 'utf-8'))
```
rwblickhan: What is `uuid1()` (i.e. why "1")? Also, does the UUID need to be converted to a string still? I imagine it returns a string already.

michaelfu1029: uuid1 is the method that generates a UUID from a host ID, sequence number, and the current time. I could also use uuid4 which generates a random UUID. I also think it needs the str() as it gives type error without. Which method would be better? uuid1 would make a duplicate if it is from the same computer and the same time, while uuid4 would make a duplicate based on random luck(around 1/2^128)

rwblickhan: Well, 2^128 is many orders of magnitude larger than the age of the universe, so I don't think we really need to worry about that 😛 As long as it works, I'd say either is fine - I don't really know enough to say either way. Perhaps @cheukyin699 has an opinion?

PR #252, opened 2019-03-09
Adding karma field to user

omar2535:
# Pull Request

## Description

Adding karma field to user

## Testing
N/A

## Ticket(s)

Sub-task of #201

### DIFF
--- a/model/user.py
+++ b/model/user.py
@@ -17,6 +17,7 @@ class User:
         self.biography = ""
         self.image_url = ""
         self.permissions_level = Permissions.member
+        self.karma = 1
 
     def get_attachment(self):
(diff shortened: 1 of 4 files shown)


model/user.py
```diff
@@ -17,6 +17,7 @@ def __init__(self, slack_id):
         self.biography = ""
         self.image_url = ""
         self.permissions_level = Permissions.member
+        self.karma = 1
```
rwblickhan: Perhaps starting with 0 karma would make more sense?

omar2535: I was thinking like reddit where everyone starts off with 1 karma, but I can roll with 0 if that's what we want.

rwblickhan: Fair enough, that makes sense.

PR #254, opened 2019-03-10
Add type annotations and static type checking

### LINKED ISSUE #23: Add type annotations to data model
Once our data model starts stabilizing a bit more, we should add type annotations from the [`typing`](https://docs.python.org/3/library/typing.html) standard library module and add type checking to our Travis builds.

cheukyin699:
# Pull Request

## Description

Add type annotations (that are sometimes somewhat complicated with the `Type[User]` and `Union` and whatnot). They are checked using `mypy` python static type checker.

Feel free to question me.

## Testing

After updating your pipenv, you can run `scripts/build_check.sh`. They should run correctly.

## Ticket(s)

Closes #23.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,18 +1,16 @@
 testing = false
+creds_path = "credentials/"
 
 [slack]
-creds_path = "credentials/slack.toml"
 bot_channel = "#rocket2"
 
 [aws]
-creds_path = "credentials/aws.toml"
 users_table = "users"
 teams_table = "teams"
 projects_table = "projects"
 region = "us-west-2"
 
 [github]
-signing_key_path = "credentials/github_signing_key.pem"
 app_id = 26418
 organization = "ubclaunchpad"
 
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,4 @@ inertia.toml
 
 # Docs generation files
 _build/
-credentials/
+/credentials/
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -1,0 +1,5 @@
+"""Pack the modules contained in the configuration directory."""
+import config.credentials
+
+Credentials = config.credentials.Credentials
+MissingCredentialsError = config.credentials.MissingCredentialsError
--- a/config/credentials.py
+++ b/config/credentials.py
@@ -1,0 +1,88 @@
+"""Contain the dictionaries of credentials for all needed services."""
+from os import listdir
+from os.path import isfile, join
+import toml
+
+
+class Credentials:
+    """Abstract away the loading of toml files."""
+
+    def __init__(self, credentials_path):
+        """
+        Collect the toml dictionaries for credentials of needed services.
+
+        :param credentials_path: path to folder that contains toml files
+                                 of credentials for all needed services
+        """
+        self.missing_cred_files = []
+        self.missing_cred_fields = {}
+
+        slack_toml_path = join(credentials_path, 'slack.toml')
+        try:
+            slack_toml = toml.load(slack_toml_path)
+            self.slack_signing_secret = self.attempt_toml_read(
+                slack_toml, 'signing_secret', 'slack')
+            self.slack_api_token = self.attempt_toml_read(
+                slack_toml, 'api_token', 'slack')
+        except IOError:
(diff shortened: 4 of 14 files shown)


tests/command/commands/user_test.py
```diff
@@ -194,26 +194,6 @@ def test_handle_edit_github(self):
         self.mock_facade.store.assert_called_once_with(user)
         self.mock_github.org_add_member.assert_called_once_with("rob")
 
-    def test_handle_edit_github_error(self):
```
rwblickhan: Why was this removed?

command/commands/token.py
```diff
@@ -18,7 +21,9 @@ class TokenCommand:
     success_msg = "This is your token:\n```\n{}\n```" \
                   "\nKeep it secret! Keep it safe!\nIt will expire at {}."
 
-    def __init__(self, db_facade, config):
+    def __init__(self,
+                 db_facade: DBFacade,
+                 config: 'TokenCommandConfig') -> None:
```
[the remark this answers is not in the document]
cheukyin699: Yes. It's for forward references. For some reason, just running mypy with some default settings allows you to take away the quotes (because it's referencing its own class). But when you run pytest with mypy plugin, it doesn't recognize its own class. Weird.

tests/command/commands/user_test.py
cheukyin699: Because it is a duplicate function (overridden immediately by the one directly below it). It seems to be a remnant of some merge conflict (woooo spooky).

command/core.py
```diff
@@ -1,26 +1,37 @@
...
 
 class Core:
     """Encapsulate methods for handling events."""
 
-    def __init__(self, db_facade, bot, gh_interface, token_config):
+    def __init__(self,
+                 db_facade: DBFacade,
+                 bot: Bot,
+                 gh_interface: GithubInterface,
+                 token_config) -> None:
```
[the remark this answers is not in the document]
cheukyin699: Yes. Let me check.

tests/command/commands/user_test.py
rwblickhan: Oh haha didn't notice that, nice catch.

command/commands/team.py
```diff
@@ -173,7 +181,8 @@ def handle(self, command, user_id):
         else:
             return self.get_help(), 200
 
-    def create_helper(self, param_list):
+    def create_helper(self,
+                      param_list: Dict[str, str]) -> ResponseTuple:
         """
```
jacqueschen1: Some of the changes here may conflict with PR for finishing team commands, not sure if you wanna push these now and we could merge later or to remove the changes for now.

interface/slack.py
```diff
@@ -7,7 +7,7 @@
 class Bot:
     """Utility class for calling Slack APIs."""
 
-    def __init__(self, sc: SlackClient) -> None:
+    def __init__(self, sc: SlackClient, slack_channel: str ="") -> None:
```
ravenCondol: fails pycodestyle  !

config.toml
```diff
@@ -1,4 +1,4 @@
-testing = false
```
rwblickhan: I don't think we should make this change - it'll probably clobber the config that's present on the server.

rwblickhan: More generally, though, I think we should probably figure out a better way to do this integration testing - we could talk about that next week though.

command/commands/token.py
```diff
@@ -30,11 +35,13 @@ def __init__(self, db_facade, config):
...
 
-    def handle(self, _command, user_id):
+    def handle(self,
+               _command: str,
+               user_id: str) -> ResponseTuple:
         """Handle request for token."""
         logging.debug("Handling token command")
         try:
-            user = self.facade.retrieve(User, user_id)
+            user = cast(User, self.facade.retrieve(User, user_id))
```
rmcreyes: I don't know much about how typing works, but would it be possible for the facade to cast the object it retrieves from the database before returning it? Saves from having to cast in every instance the facade returns something.

ravenCondol:
WHAT DID I JUST DO

PR #256, opened 2019-03-11
Clean up imports

### LINKED ISSUE #255: __init__ files should have class aliases

Right now we are doing the following, which is very redundant:

```python
from model.user import User
from model.team import Team
from model.project import Project
```

When most of the code bases I've seen this happens

```python
from model import User, Team, Project
```

For `model/__init__.py`:

```python
import model.user
import model.team
import model.project

User = model.user.User
Team = model.team.Team
Project = model.project.Project
```

rwblickhan:
# Pull Request

## Description

Clean up the imports, in particular by adding reexports for model classes as requested in #255.

## Ticket(s)

Closes #255.


command/core.py
```diff
@@ -1,11 +1,12 @@
 """Calls the appropriate handler depending on the event data."""
-from command.commands.user import UserCommand
-from command.commands.token import TokenCommand
 import command.util as util
-from model.user import User
-from interface.slack import SlackAPIError
 import logging
+
+from command.commands.user import UserCommand
+from command.commands.token import TokenCommand
```
cheukyin699: Might wanna change this to `from command.commands import UserCommand, TokenComand`, or something even simpler.

tests/webhook/webhook_test.py
```diff
@@ -1,10 +1,10 @@
 """Test the webhook handler."""
 import pytest
+
+from db.facade import DBFacade
```
cheukyin699: For db as well. `from db import DBFacade`

command/core.py
rwblickhan: I considered that, but the `TokenCommand` also exports the command configuration class, and I imagine other commands might do that in the future as well, so `__init__` for `command.commands` could become quite cluttered. But if we're okay with that, I can add that.

tests/webhook/webhook_test.py
rwblickhan: Sure.

factory/__init__.py
```diff
@@ -8,7 +8,7 @@
 from command.core import Core
 from command.commands.token import TokenCommandConfig
 from datetime import timedelta
-from db.facade import DBFacade
+from db import DBFacade
 from db.dynamodb import DynamoDB
```
cheukyin699: and one for DynamoDB as well?

rwblickhan: I'm not sure we want to do that. The DynamoDB class is supposed to be an implementation detail, so I'm not convinced we want to reexport it at the `db` level.

command/commands/team.py
```diff
@@ -211,5 +212,5 @@ def create_helper(self,
             return msg, 200
         except GithubAPIException as e:
             logging.error("team creation unsuccessful")
-            return "Team creation unsuccessful with the following error"\
+            return "Team creation unsuccessful with the following error" \
```
CoffeeTableEspresso:
```suggestion
return f"Team creation unsuccessful with the following error: {e.data}", 200
```

rwblickhan: I would make this change but I believe a number of these functions have similar issues. Do you mind making a new ticket to fix them, so we can do them all in one sweep?

PR #257, opened 2019-03-12
Deployment

### LINKED ISSUE #104: Write deployment documentation
We should write some documentation explaining how we've set up with Docker Compose, Inertia, etc.

rwblickhan:
# Pull Request

## Description

Our previous deployment (in `docker-compose.yml`) was somewhat iffy - it was reliant on a pair of containers, for Nginx and Let's Encrypt, that exposed the Docker socket (a very bad idea, according to @bobheadxi) and also tended to fail when pull requests were merged. The new configuration should work much better.

I've also finally added the deployment documentation requested in #104.

## Ticket(s)

Closes #104


docker-compose.yml
```diff
@@ -1,30 +1,22 @@
...
     rocket2:
         build:
             context: .
             dockerfile: Dockerfile
-        environment:
-            VIRTUAL_HOST: "${VIRTUAL_HOST}"
-            LETSENCRYPT_HOST: "${LETSENCRYPT_HOST}"
-            LETSENCRYPT_EMAIL: "${LETSENCRYPT_EMAIL}"
+        ports:
+            - 5000:5000
```
bobheadxi: 🚀

server/server.py
```diff
@@ -48,6 +48,7 @@
 app = Flask(__name__)
 # HTTP security header middleware for Flask
 talisman = Talisman(app)
+talisman.force_https = False
```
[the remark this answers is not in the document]
rwblickhan: Yeah, it was tripping up Nginx when it made requests to the actual Rocket server.

docs/Deployment.md
```diff
@@ -1,6 +1,44 @@
...
+# Deployment
 
-## Travis CI
+## Deployment Process
+
+### SSL
+
+Before deploying for the first time, you must set up SSL and configuration for
+Nginx, which we are using as a proxy server. This can be done by running the
+[`setup_deploy.sh`](../scripts/setup_deploy.sh) script. This runs the official
```
cheukyin699: This does not work as you think it should (see the built docs [here][built-docs]). The links are valid for within GitHub only, but invalid once built at RTD. Consider a hard link or removing them??? Same with the ones below.

[built-docs]: https://rocket2.readthedocs.io/en/deployment/docs/Deployment.html

rwblickhan: Sad. I've just gone ahead and removed them - I think it's obvious enough where they come from.

PR #247, opened 2019-03-02
Michaelfu1029/#166 create membership GitHub webhooks

michaelfu1029:
# Pull Request

## Description

Implemented the membership webhook and updated an aspect that alerts us if there is something wrong with the database(multiple stack IDs with one github username)

## Testing

No additional changes required for testing setup

## Ticket(s)

Closes #166

PR #258, opened 2019-03-13
Use formatted strings (f-strings)everywhere

### LINKED ISSUE #246: Use formatted strings everywhere
Python 3.6 onwards supports something called `f-strings`. Instead of `"x is " + str(x)`, you can write `f"x is {x}"`. f-strings use the same formatting syntax as the existing `format`, but are much cleaner.

I suggest using f-strings throughout our codebase, and in the future for new code.

michaelfu1029:
# Pull Request

## Description
Changing string literals to f-strings. Did not change team.py as it is currerently stubs, and webhook.py, as some functions are not merged yet(fixed in the other PR #247 )

## Ticket(s)

Closes #246


interface/slack.py
```diff
@@ -13,42 +13,42 @@ def __init__(self, sc):
...
             "chat.postMessage",
             channel=channel_name,
             attachments=attachments,
             text=message
         )
         if 'ok' not in response:
-            logging.error("Message to channel {} failed with error: {}".
-                          format(channel_name, response['error']))
+            logging.error(f"Message to channel {channel_name} failed with "
+                          f"error: { response['error']}")
```
cheukyin699: Space between left bracket and `r` where there is no space between the right bracket: consistency

michaelfu1029: fixed

command/commands/user.py
```diff
@@ -183,8 +183,8 @@ def edit_helper(self, user_id, param_list):
                 self.github.org_add_member(param_list["github"])
                 edited_user.github_username = param_list["github"]
             except GithubAPIException as e:
-                msg = "\nError adding user {} to GitHub organization".format(
-                    param_list['github'])
+                msg = (f"\nError adding user {param_list['github']} to GitHub"
```
rwblickhan: I think the parentheses here can be removed.

interface/slack.py
```diff
@@ -13,42 +13,42 @@ def __init__(self, sc, slack_channel=""):
...
+        logging.debug(f"Sending direct message to {slack_user_id}")
         response = self.sc.api_call(
             "chat.postMessage",
             channel=slack_user_id,
             text=message
         )
-        if 'ok' not in response or not response['ok']:
-            logging.error("Direct message to {} failed with error: {}".
-                          format(slack_user_id, response['error']))
+        if 'ok' not in response:
```
ravenCondol: why change ?

interface/slack.py
```diff
@@ -13,42 +13,42 @@ def __init__(self, sc, slack_channel=""):
...
         response = self.sc.api_call(
             "chat.postMessage",
             channel=channel_name,
             attachments=attachments,
             text=message
         )
-        if 'ok' not in response or not response['ok']:
-            logging.error("Message to channel {} failed with error: {}".
-                          format(channel_name, response['error']))
+        if 'ok' not in response:
```
ravenCondol: here too, think the or is important here

PR #259, opened 2019-03-14
refactor credentials loading

### LINKED ISSUE #241: Refactor credentials loading
Right now our credentials loading is somewhat haphazard. We should consider refactoring this so that it's more standardized.

rmcreyes:
# Pull Request

## Description

Created a class that took the path of the `credentials` folder and performed all the `toml.load()`ing needed to avoid doing so in the startup factory. Replaced all instances of loading `toml` files with just using the `Credentials` object's attributes.

## Testing

## Ticket(s)

Closes #241 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,4 @@ inertia.toml
 
 # Docs generation files
 _build/
-credentials/
+/credentials/
--- a/config.toml
+++ b/config.toml
@@ -2,18 +2,15 @@ testing = false
 creds_path = "credentials/"
 
 [slack]
-creds_path = "credentials/slack.toml"
 bot_channel = "#rocket2"
 
 [aws]
-creds_path = "credentials/aws.toml"
 users_table = "users"
 teams_table = "teams"
 projects_table = "projects"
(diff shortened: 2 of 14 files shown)


rmcreyes:
@ubclaunchpad/brussel-sprouts could use some advice on how to assign the fields more elegantly, as I think if we want to use explicit declarations of attributes, a lot of repetition may need to be involved. A pretty hacky way exists where I can use `exec()` to use iteration for declaring attributes, like:
```
exec("self.{} = {}".format(credential_attribute_name, credential_attribute_value)
```
but that looks kind of jank

.gitignore
```diff
@@ -17,4 +17,4 @@ inertia.toml
 
 # Docs generation files
 _build/
-credentials/
+/credentials/
```
cheukyin699: wait why?

rmcreyes: Apparently makes it so that it only ignores the `credentials` directory in the root directory. Needed to make another one (`tests/credentials`) to test with actual `toml` files. (don't know how good this idea actually is ehe)

PR #262, opened 2019-03-17
readded csrf security

### LINKED ISSUE #224: Add Flask security middleware

rmcreyes:
# Pull Request

## Description

Readded Seasurf middleware for cross-site request forgery. The problem we had the first time was that because all requests to the server are being done by third party applications like Slack and GitHub, there was no csrf token for Seasurf to validate, and Seasurf would then reject requests to any of our endpoints. This can be worked around by using the `exempt` decorator, but since we eventually plan to allow requests to the endpoints without using third party applications, we will need to take into consideration the authentication of csrf tokens for whenever we plan to do that.

## Testing

It shouldn't reject requests when you try to run rocket now

## Ticket(s)

Closes #224

### DIFF
--- a/server/server.py
+++ b/server/server.py
@@ -52,6 +52,8 @@ app = Flask(__name__)
 # HTTP security header middleware for Flask
 talisman = Talisman(app)
 talisman.force_https = False
+# cross-site request forgery security middleware for Flask
+csrf = SeaSurf(app)
 config = toml.load('config.toml')
 credentials = Credentials(config['creds_path'])
 core = make_core(config, credentials)
@@ -64,6 +66,7 @@ slack_events_adapter = SlackEventAdapter(slack_signing_secret,
                                          "/slack/events", app)
 
 
+@csrf.exempt
 @app.route('/')
 def check():
     """Display a Rocket status image."""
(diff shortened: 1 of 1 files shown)


server/server.py
```diff
@@ -64,13 +66,15 @@
                                          "/slack/events", app)
 
 
+@csrf.exempt
 @app.route('/')
```
rwblickhan: This endpoint doesn't need to be exempted, right? It'll only be accessed directly through a browser, not through redirects from Slack/Github/etc.

rmcreyes: I can't seem to reach the endpoint if it isn't exempted cause it still seems to attempt verifying a csrf token. After doing a bit more research I'm getting the vibe that protection against csrf attacks are more for applications that rely on forms and keep track of sessions, while currently we just provide a REST application. Might need another opinion on this too, but even then I don't think the root endpoint does anything to expose a security vulnerability.

rwblickhan: Yeah, that's true. In that case, perhaps we just shouldn't bother with this, then?

rmcreyes: Maybe, unless we decide that we want to interact with rocket through forms later on, but I'd imagine that wouldn't be in the near future.

rwblickhan: True - that's very much not in the plans right now (though if we ever add @cheukyin699's idea for an admin console...)

PR #263, opened 2019-03-18
 Add type annotations and static type checking

### LINKED ISSUE #23: Add type annotations to data model

cheukyin699:
# Pull Request

## Description

Add type annotations (that are sometimes somewhat complicated with the `Type[User]` and `Union` and whatnot). They are checked using `mypy` python static type checker.

Feel free to question me.

## Testing

After updating your pipenv, you can run `scripts/build_check.sh`. They should run correctly.

## Ticket(s)

Closes #23.

### DIFF
--- a/command/commands/token.py
+++ b/command/commands/token.py
@@ -2,9 +2,12 @@
 import jwt
 import logging
 
-from datetime import datetime
+from typing import cast
+from datetime import datetime, timedelta
+from command import ResponseTuple
 from model.permissions import Permissions
 from model.user import User
+from db.facade import DBFacade
 
 
 class TokenCommand:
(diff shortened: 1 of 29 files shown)


command/commands/token.py
```diff
@@ -30,11 +35,13 @@ def __init__(self, db_facade, config):
...
 
-    def handle(self, _command, user_id):
+    def handle(self,
+               _command: str,
+               user_id: str) -> ResponseTuple:
         """Handle request for token."""
         logging.debug("Handling token command")
         try:
-            user = self.facade.retrieve(User, user_id)
+            user = cast(User, self.facade.retrieve(User, user_id))
```
rmcreyes: Mentioned this in the last pr, but could the casting be done at the `dynamodb` or `facade` level of `retrieve()`?

cheukyin699: I was wondering how the generics would work but then i figured it out and now you no longer have to explicitly cast it to something.

QED
YAY

PR #268, opened 2019-03-22
Secure GitHub webhooks

### LINKED ISSUE #261: Secure Github webhooks

Our webhooks technically accept any POST requests. We should authenticate them so we only accept actual Github POST requests.

See the ["Securing your webhooks"](https://developer.github.com/webhooks/securing/) page on the Github API.

rmcreyes:
# Pull Request

## Description

Added hash signature verification to so endpoints would only properly serve requests sent by GitHub rather than any HTTP request.

## Testing

It will work if you are listening to a webhook event and have set a webhook secret. In my opinion adding the documentation of the process of setting this up should be part of @michaelfu1029 's ticket when he refactors webhooks to only use one endpoint with the GitHub app.

## Ticket(s)

Closes #261 


webhook/webhook.py
```diff
@@ -1,18 +1,41 @@
...
+        :param xhub_signature: Hashed signature to validate
+        :return: Return True if the signature is valid, False otherwise
+        """
+        h = hmac.new(bytes(self.__secret, encoding='utf8'),
+                     request_body, hashlib.sha1)
+        verified = hmac.compare_digest(
+            bytes("sha1=" + h.hexdigest(), encoding='utf8'),
+            bytes(xhub_signature, encoding='utf8'))
+        if verified:
+            logging.info("Webhook signature verified")
```
rwblickhan: I think this log specifically should be `debug` level - we expect most of the signatures to succeed 😛

webhook/webhook.py
```diff
@@ -1,18 +1,41 @@
...
+        else:
+            logging.warning("Webhook signature not verified")
```
rwblickhan: I think it would be valuable to specify that this is (supposed to be) a _Github_ webhook. Also, it might be useful to include some more details about it - e.g. maybe include the signature that failed to verify?

server/server.py
```diff
@@ -1,113 +1,125 @@
...
+
+@app.route('/webhook/organization', methods=['POST'])
+def handle_organization_webhook():
+    """Handle GitHub organization webhooks."""
+    logging.info("organization webhook triggered")
+    xhub_signature = request.headers.get('X-Hub-Signature')
+    request_data = request.get_data()
+    request_json = request.get_json()
+    logging.debug(f"organization payload: {str(request_json)}")
+    if webhook_handler.verify_hash(request_data, xhub_signature) is True:
```
cheukyin699:
```suggestion
    if webhook_handler.verify_hash(request_data, xhub_signature):
```

server/server.py
```diff
@@ -1,113 +1,125 @@
...
+    """Handle GitHub team webhooks."""
+    logging.info("team webhook triggered")
+    xhub_signature = request.headers.get('X-Hub-Signature')
+    request_data = request.get_data()
+    request_json = request.get_json()
+    logging.debug(f"team payload: {str(request.get_json())}")
+    if webhook_handler.verify_hash(request_data, xhub_signature) is True:
+        msg = webhook_handler.handle_team_event(request.get_json())
+    else:
+        msg = "Hashed signature is not valid", 403
```
cheukyin699: This hash verification looks like it could be abstracted out into a separate function.

webhook/webhook.py
rmcreyes: done

webhook/webhook.py
rmcreyes: and done?

server/server.py
rmcreyes: Not sure what exactly this means but see what I did I guess?

PR #269, opened 2019-03-22
Jacqueschen1/142 team commands

### LINKED ISSUE #142: Implement team commands

jacqueschen1:
# Pull Request

## Description

Implemented all commands except `team list` (had some trouble with json), and added tests. There are a few strange bits that need to be looked over, such as deciding if certain commands should be interacting with the database vs GitHub.

## Testing

## Ticket(s)

Closes #142 


command/util.py
```diff
@@ -40,3 +41,21 @@ def escaped_id_to_id(s):
...
                   r"\1",
                   s)
+
+
+def check_credentials(user, team):
+        """
+        Check if given user is admin or team lead.
+
+        If team is specified and user is not admin, checks if user is
+        team lead in team.
```
cheukyin699: Should also specify what would happen if team is **not** specified.

model/team.py
```diff
@@ -17,6 +17,40 @@ def __init__(self, github_team_id, github_team_name, display_name):
...
+        text_pairs = [
+            ('Github ID', self.github_team_id),
+            ('Github Team Name', self.github_team_name),
+            ('Display Name', self.display_name),
+            ('Platform', self.platform),
+            ('Team Leads' '\n'.join(self.team_leads)),
+            ('Members', '\n'.join(self.members))
+        ]
+        fields = [{'title': t, 'value': v if v else 'n/a', 'short': True}
+                  for t, v in text_pairs]
```
cheukyin699: Since the value we are checking is always initialized, we are checking to see if it is an empty string or not. Thus, we can simply remove the tertiary operators in `get_attachment(self)`, and combine the 2 functions into 1 function.

command/commands/team.py
```diff
@@ -182,25 +227,219 @@ def create_helper(self, param_list):
...
+                team.platform = param_list['platform']
             if param_list["channel"] is not None:
-                msg += "add channel"
-                # stub: cannot finish until teamMember PR is pushed
-            self.facade.store_team(team)
+                msg += "added channel, "
+                for member_id in self.sc.get_channel_users(
+                        param_list["channel"]):
+                    member = self.facade.retrieve(User, member_id)
+                    self.gh.add_team_member(member.github_username, team_id)
```
cheukyin699: Should you also add the team member to the Team object, or is this done with a webhook?

command/commands/team.py
```diff
@@ -182,25 +227,219 @@ def create_helper(self, param_list):
...
+                    self.gh.add_team_member(lead_user.github_username, team_id)
+            else:
+                team.add_team_lead(command_user.github_id)
+
+            self.facade.store(team)
             return msg, 200
         except GithubAPIException as e:
             logging.error("team creation unsuccessful")
-            return "Team creation unsuccessful with the following error"\
+            return "Team creation unsuccessful with the following error: "\
```
cheukyin699:
```suggestion
            return f"Team creation unsuccessful with the following error: {e.data}", 200
```

Though you might wanna split it up a little bit.

command/commands/team.py
cheukyin699: Also may need to handle the case where the user's Github username isn't entered, and come up with a list of names that don't have Github usernames

command/commands/team.py
```diff
@@ -182,25 +227,219 @@ def create_helper(self, param_list):
...
+        :param user_id: Slack ID of user who called command
+        :return: return error message if user added unsuccessfully
+                 or if user has insufficient permission level,
+                 otherwise returns success message
+        """
+        try:
+            command_user = self.facade.retrieve(User, user_id)
+            team = self.facade.retrieve(Team, param_list['team_name'])
+            if not check_credentials(command_user, team):
+                return self.permission_error, 200
```
cheukyin699: This 4 line block of code is so commonly used that I think you should add a function decorator to reduce the amount of typing (and so that it looks prettier).

command/commands/team.py
```diff
@@ -182,25 +227,219 @@ def create_helper(self, param_list):
...
+
+            user = self.facade.retrieve(User, param_list['slack_id'])
+            if not self.gh.has_team_member(user.github_username,
+                                           team.github_team_id):
+                return "User not in team!", 200
+            team.discard_member(user.github_id)
+            team.remove_team_lead(user.github_id)
```
cheukyin699: Uh okay.

model/team.py
jacqueschen1: will try to see if it works

command/commands/team.py
jacqueschen1: Github username is the only way to add members using pyGithub, users should reaaaalllly have their usernames specified :p

jacqueschen1: Oh good catch, might as well

rwblickhan:
Looks like there's quite a few style issues - friendly reminder that you can run `pipenv run pycodestyle .` or `./scripts/build_check.sh` to run those locally 😃

ravenCondol:
there are some that dont seem to show up on the local ones but do on travis though

command/commands/team.py
```diff
@@ -2,16 +2,21 @@
...
 
 
 class TeamCommand:
     """Represent Team Command Parser."""
 
     command_name = "team"
     desc = "for dealing with " + command_name + "s"
-    permission_error = "Insufficient permission level to run command!"
+    permission_error = "You do not have the sufficient " \
+                       "permission level for this command!"
```
rwblickhan: I think it would flow better if this was changed to `"You do not have sufficient permissions for this command!"`

command/commands/team.py
```diff
@@ -2,16 +2,21 @@
...
+    lookup_error = "Lookup error! User not found!"
```
rwblickhan: I think this would look a little bit nicer as `"Lookup error: user not found!"`

command/commands/team.py
```diff
@@ -102,6 +110,17 @@ def init_subparsers(self):
                                  help="Display name the team should have.")
         parser_edit.add_argument("--platform", type=str, action='store',
                                  help="Platform the team should have.")
+
+        """Parser for lead command."""
+        parser_lead = subparsers.add_parser("lead")
+        parser_lead.set_defaults(which='lead',
+                                 help="Edit leads of specified team.")
+        parser_lead.add_argument("team_name", type=str, action='store',
```
rwblickhan: I believe we're changing all these underscores to hyphens (@cheukyin699?).

jacqueschen1: oh boy

jacqueschen1: so I tried replacing `team_name` with `team-name` but I get an error whenever I call `args.team-name` where it says `name` is an unresolved reference

ravenCondol: no idea if this helps but https://stackoverflow.com/questions/12834785/having-options-in-argparse-with-a-dash

jacqueschen1: Should i just change the `team_name` in the add_argument lines and none of the others?

cheukyin699: you should keep that as `team_name` and instead add parameter `metavar='team-name'`. That way, you can keep using `arg.team_name` while also displaying `team-name`.

cheukyin699: Above was taken from link that @ravenCondol found.

command/commands/team.py
```diff
@@ -2,24 +2,29 @@
 import argparse
 import logging
 import shlex
+from flask import jsonify
 from interface.github import GithubAPIException
 from model.team import Team
+from model.user import User
```
rmcreyes:
```suggestion
from model import Team, User
```

command/util.py
```diff
@@ -40,3 +41,22 @@ def escaped_id_to_id(s):
     return re.sub(r"<@(\w+)\|[^>]+>",
                   r"\1",
                   s)
+
+
+def check_credentials(user, team):
```
ravenCondol: could this be called check_pawmissions instead ? 😸 i think itd be clearer imo since we use credentials to mean other things too

jacqueschen1: Yah sure

command/commands/team.py
cheukyin699: There are a few other places where this occurs. Search for string `from model\.`.

PR #274, opened 2019-03-23
updated the doc to include the checkbox for escaping

### LINKED ISSUE #225: Update dev docs to include option to escape usernames, channels, and links
Because of #219 , we check for all escaped strings and convert them into either user IDs or channels IDs. You can turn this option on with a flick of a checkbox (mentioned in https://github.com/ubclaunchpad/rocket2/issues/212#issuecomment-462087825). Please update the local development documentation to include this setting. It will be greatly appreciated.

michaelfu1029:
# Pull Request

## Description
Updated the documentation for local developer guide

## Ticket(s)

Closes #225


docs/LocalDevelopmentGuide.md
```diff
@@ -175,6 +175,9 @@ Request URL, submit the ngrok HTTPS URL with `/slack/commands` appended to the
 end. For the actual command, anything will work, though the final app will use
 `/rocket`.
 
+Select the checkbox titled "Escape channels, users, and links sent to your app"
```
omar2535: Shouldn't the command be under a new section like 7.4?

cheukyin699: It's in the same dialog box, so probably most likely no.

docs/LocalDevelopmentGuide.md
```diff
@@ -175,6 +175,9 @@ Request URL, submit the ngrok HTTPS URL with `/slack/commands` appended to the
...
+if you wish to escape usernames, channels, and links.
```
cheukyin699: Maybe make a note that you must check this box to have everything work properly

rwblickhan: Agreed - this is a requirement for the app, and some things will break if this option isn't enabled.

PR #277, opened 2019-03-27
replaced underscores with hyphens for user command arguments

### LINKED ISSUE #270: Hyphen/Underscore delimiting words in argparse

Replace the commands that have underscores with hyphens. Argparse takes care of the hyphens by converting them into underscores, but we will just see the hyphens in the help text, which helps with consistency.

Please replace wherever there are underscores in declaring the parsers with hyphens, and replace the documentations. Remember to change the tests!

Should be pretty locally contained

rmcreyes:
# Pull Request

## Description

Replaced underscores with hyphens in user command arguments.

## Testing

## Ticket(s)

Closes #270 


command/commands/user.py
```diff
@@ -43,7 +43,8 @@ def init_subparsers(self) -> _SubParsersAction:
         parser_view.set_defaults(which="view",
                                  help="View information about a given user.")
         parser_view. \
-            add_argument("--slack_id", type=str, action='store',
+            add_argument("--slack-id", metavar="SLACK-ID",
```
cheukyin699: I am reading the docs and it looks like there is simply no mention of `--slack-id`, even though the code references it. @rwblickhan do you think we should remove the two dashes before `--slack-id` so that it is treated as a variable instead of a flag?

Docs in question: https://rocket2.readthedocs.io/en/latest/docs/UserCommands.html#view

rwblickhan: Hmm, yes, seems the implementation diverged a bit from the spec. That being said, I actually like the implementation better - I can do `/rocket user view` just to view my own profile, or `/rocket user view --slack-id @user` to specify another one. If there's not a good argument against, I would say the docs should be changed, instead.

rwblickhan: What does `metavar` do?

command/commands/user.py
```diff
@@ -59,7 +60,8 @@ def init_subparsers(self) -> _SubParsersAction:
         parser_delete.set_defaults(which="delete",
                                    help="(Admin only) permanently delete "
                                         "member's profile.")
-        parser_delete.add_argument("slack_id", type=str, action='store',
+        parser_delete.add_argument("slack_id", metavar="slack-id",
```
rwblickhan: This should also be changed from an underscore to a hyphen, right? Why is `metavar` different here from above?

command/commands/user.py
rmcreyes: the metavar in this case represents the value that would be replaced in `/rocket user view --slack-id SLACK-ID`

command/commands/user.py
rmcreyes: the metavar in this case represents the value that would replace the positional argument in `/rocket user delete slack-id`. For some reason `argparse` only converts hyphens to underscores for optional arguments but not positional arguments.

PR #282, opened 2019-03-29
Raven c/#237 rocket refresh team command

### LINKED ISSUE #237: Add `/rocket refresh`
Add `/rocket refresh` or similar command.

There's a (small) possibility that teams will be created (or so on) in Github but the webhook will not be delivered. Add a command (or possibly something that runs automatically every so often) to grab the latest details from Github and write them into the Rocket database, as necessary.

ravenCondol:
# Pull Request
`/rocket team refresh` ensures local and github teams are synced. Upon command invocation, bot tells you which teams were deleted/edited/added (up to a limit of 100 due to slack api limitations).

If a team exists in both already, a weaker check for equality (one based on team name and members ONLY) is done, updating both if the check fails. This can be changed if needed.

~~Kinda off thing I noticed is how, now, team command works with actual github team objects, rather than just models of them.~~ Hopefully resolved.

## Ticket(s)

Closes #237


command/commands/team.py
```diff
@@ -214,3 +215,58 @@ def create_helper(self,
...
+                 otherwise returns success messages with # of teams changed
+        """
+        num_changed = 0
+        num_added = 0
+        num_deleted = 0
+        modified = []
+        try:
+            local_teams = self.facade.query(Team)
+            remote_teams = self.gh.org_get_teams()
+            local_ids = dict((Team.to_dict(team).get('github_team_id'), team)
```
cheukyin699:
```suggestion
            local_ids = dict((team.github_team_id, team) for team in local_teams)
```

model/team.py
```diff
@@ -21,6 +21,24 @@ def __init__(self,
...
 
+    def get_attachment(self):
+        """Return slack-formatted attachment (dictionary) for team."""
+        text_pairs = [
+            ('Github ID', self.github_team_id),
+            ('Github Team Name', self.github_team_name),
+            ('Display Name', self.display_name),
+            ('Platform', self.platform),
+            ('Team Leads', '\n'.join(self.team_leads)
+                           if self.team_leads != set() else 'n/a'),
```
cheukyin699: You can remove the ternary if statement(s) since line 36 will check to see if any of the fields are empty or not, and replace them with `n/a`.

tests/command/commands/team_test.py
```diff
@@ -471,3 +471,94 @@ def test_handle_edit_lookup_error(self):
...
+        self.gh.org_get_teams.return_value = [team]
+        attach = team.get_attachment()
+        attach2 = team2.get_attachment()
+
+        status = f"0 teams changed, " \
+            f"1 added, " \
+            f"1 deleted. Wonderful."
+        with self.app.app_context():
+            resp, code = self.testcommand.handle("team refresh", user)
+            expect = json.loads(jsonify({'attachments': [attach2, attach],
```
ravenCondol: this test fails if i write `'attachments': [attach, attach2]` (aka i change the order of the list). this was weird, since i thought that assertDictEquals ignores order of dictionary keys, but i guess it does not ignore order for dictionary values

command/commands/team.py
```diff
@@ -476,3 +486,59 @@ def delete_helper(self, team_name, user_id) -> ResponseTuple:
...
+        modified = []
+        try:
+            command_user = self.facade.retrieve(User, user_id)
+            if not check_permissions(command_user, None):
+                return self.permission_error, 200
+            local_teams = self.facade.query(Team)
+            remote_teams = self.gh.org_get_teams()
+            local_ids = dict((team.github_team_id, team)
+                             for team in local_teams)
+            remote_ids = dict((team.github_team_id, team)
```
jacqueschen1: `org_get_teams()` returns a list of `Team` objects that are from PyGithub, so they aren't using our team implementation. You should call `team.id` instead of `team.github_team_id`

jacqueschen1: https://pygithub.readthedocs.io/en/latest/github_objects/Team.html?highlight=team

command/commands/team.py
```diff
@@ -476,3 +486,59 @@ def delete_helper(self, team_name, user_id) -> ResponseTuple:
...
+            for local_id in local_ids:
+                if local_id not in remote_ids:
+                    self.gh.org_delete_team(local_id)
+                    num_deleted += 1
+                    modified.append(local_ids[local_id].get_attachment())
+
+            # add teams to db that are in github but not in local database,
+            for remote_id in remote_ids:
+                if remote_id not in local_ids:
+                    self.facade.store(remote_ids[remote_id])
```
jacqueschen1: You should create a new team object here and store it, since the `remote_id` teams aren't Teams from our model, but PyGithub Team objects. Sorry for the confusion :/

ravenCondol: Ty! Done

command/commands/team.py
ravenCondol: yeet

tests/command/commands/team_test.py
```diff
@@ -466,3 +467,105 @@ def test_handle_edit_lookup_error(self):
...
+        team2 = Team("OTEAM", "other team2", "ios")
+
+        self.db.retrieve.return_value = test_user
+        self.db.query.return_value = [team, team2]
+
+        # In this case, github has an updated version of team!
+        remote_team_update = mock.Mock(GithubTeam.Team)
+        remote_team_update.id = team_update.github_team_id
+        remote_team_update.name = team_update.github_team_name
+
```
ravenCondol: One could be more stringent: 
```suggestion
remote_member1 = mock.Mock(github.NamedUser.NamedUser)
remote_member1.setFields(member1)
remote_team_update.members = PaginatedList(remote_member1)
```
here, but it feels unnecessary

command/commands/team.py
```diff
@@ -476,3 +486,67 @@ def delete_helper(self, team_name, user_id) -> ResponseTuple:
...
+        modified = []
+        try:
+            command_user = self.facade.retrieve(User, user_id)
+            if not check_permissions(command_user, None):
+                return self.permission_error, 200
+            local_teams: List[Team] = self.facade.query(Team)
+            remote_teams: List[GithubTeam.Team] = self.gh.org_get_teams()
+            local_ids = dict((team.github_team_id, team)
+                             for team in local_teams)
+            remote_ids = dict((str(team.id),
```
rmcreyes: I hate to be the picky one but I don't like the idea of calling a dictionary by something expected to be its key ._.

command/commands/team.py
```diff
@@ -476,3 +486,67 @@ def delete_helper(self, team_name, user_id) -> ResponseTuple:
...
+            return "Refresh teams was unsuccessful with " \
+                   f"the following error: {e.data}", 200
+        except LookupError:
+            logging.error("team refresh unsuccessful due to lookup error")
+            return self.lookup_error, 200
+        status = f"{num_changed} teams changed, " \
+            f"{num_added} added, " \
+            f"{num_deleted} deleted. Wonderful."
+        ret = {'attachments': modified, 'text': status}
+        return jsonify(ret), 200
```
rmcreyes: @cheukyin699 so in what cases do we not return error code 200?

cheukyin699: We don't. We return nothing but 200s, according to [slack documentation][docs] on the slash command.

[docs]: https://api.slack.com/slash-commands#responding_with_errors

command/commands/team.py
cheukyin699: What do you mean by that?

command/commands/team.py
rmcreyes: hmm ok

ravenCondol: I think I understand. What should I call it instead?

command/commands/team.py
ravenCondol: I think I understand. What should I call it instead?

command/commands/team.py
```diff
@@ -476,3 +486,67 @@ def delete_helper(self, team_name, user_id) -> ResponseTuple:
...
+        try:
+            command_user = self.facade.retrieve(User, user_id)
+            if not check_permissions(command_user, None):
+                return self.permission_error, 200
+            local_teams: List[Team] = self.facade.query(Team)
+            remote_teams: List[GithubTeam.Team] = self.gh.org_get_teams()
+            local_ids = dict((team.github_team_id, team)
+                             for team in local_teams)
+            remote_ids = dict((str(team.id),
+                               Team(str(team.id), team.name, ""))
```
cheukyin699: As far as I'm aware, this does not populate the `members` attribute. You must do so, maybe with a small helper function. Some kind of utility function `convert_GithubTeam_to_DBTeam` or something shorter.

Because, later, you will find that there are no members in the Team (meaning that it is modified), and you will be replacing the set of members (probably quite accurate) with an empty set, effectively Thanos-snapping them out of existence.

Please nerf.

ravenCondol: ahk, I knew not having teams containing members was going to bite me. i will get on it promptly, and by promptly i mean sometime between now and the heat death of the universe

cheukyin699: Very accurate; just not very precise.:)

ravenCondol: should I just make the github interface's get_teams return actual teams we can work with, rather than github ones?

cheukyin699: yes

command/commands/team.py
rmcreyes: `local_teams_dict` and `remote_teams_dict` maybe?

cheukyin699: ohhh I get it now

interface/github.py
```diff
@@ -111,7 +112,9 @@ def org_get_teams(self):
         team_array = []
         for team in teams:
             # convert PaginatedList to List
-            team_array.append(team)
+            team_model = ModelTeam(team.id, team.name, "")
+            team_model.members = set(self.list_team_members(team.id))
```
cheukyin699: I should probably add type annotations to `interface/*.py` files, but...

`list_team_members` returns a list of the team members, not the member's github ids. You would probably need to change the function to return that, or select it using a map lambda.

ravenCondol: ooooo okok

ravenCondol: would something like this work? 
```suggestion
            team_model.members = set(user.id
                                     for user in
                                     self.list_team_members(team.id))
```

cheukyin699: Yes. Most likely.

ravenCondol:
tfw you procrastinate on studying for finals by doing work ._.

cheukyin699:
@rwblickhan how do i delete someone else's comment

PR #286, opened 2019-03-30
Fix member webhook name and query strings

### LINKED ISSUE #271: Membership webhook set up to receive incorrect events
**Describe the bug**

The `handle_membership_event()` in `webhook/webhook.py` is set up to respond to event actions called `member_added`, `member_removed`, and `member_invited`, but a look at the GitHub webhook documentation says that the membership event is supposed to fire actions called `added` and `removed`. 

**To Reproduce**

The webhook will not be handled correctly if manually tested.

**Expected behavior**

The `handle_membership_event()` should evaluate the action based on the `added` and `removed` action names. See https://developer.github.com/v3/activity/events/types/#membershipevent. Also wouldn't hurt to actually use the function in an endpoint in `server.py`.

cheukyin699:
# Pull Request

## Description

- [x] Fix member webhook action names
- [x] Fix query strings used in webhooks

## Testing

Note: this has not been tested on github; don't know if this works.

## Ticket(s)

Closes #271

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-testing = true
+testing = false
 creds_path = "credentials/"
 
 [slack]
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-testing = false
+testing = true
 creds_path = "credentials/"
 
 [slack]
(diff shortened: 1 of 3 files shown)


config.toml
```diff
@@ -1,4 +1,4 @@
-testing = false
+testing = true
```
ravenCondol: i keep seeing this flicking back and forth between different branches. which one is it supposed to be ? ?

cheukyin699: bad habits sorry

PR #302, opened 2019-05-25
Make scripts to set up and validate your credentials

### LINKED ISSUE #285: Add script to set up credentials

Add a script that copies credentials to the correct place.

We have a lot of required credentials. It would be nice to have a script that validates that all required credentials are in place, or allows you to one-by-one copy them into the credentials directory.

cheukyin699:
# Pull Request

## Description

The script will set up and validate your credentials. Not much more to say. Everything in your `credentials` folder will be filled, and the variables that are necessary should, at the very least, exist. This script also `touch`es the `signing_key` that you get from Github.

## Testing

Save your credentials into a secure location and remove your existing credentials. Then run the script. You should find a bunch of credential files there.

## Ticket(s)

Closes #285

### DIFF
--- a/scripts/setup_credentials.sh
+++ b/scripts/setup_credentials.sh
@@ -10,42 +10,58 @@ fi
 problems=0
 CREDS=$(grep creds_path "$CONF" | cut -d ' ' -f 3 | tr -d '"')
 SIGNING=$(grep signing_key_path "$CONF" | cut -d ' ' -f 3 | tr -d '"')
+declare -A FILESCHANGED         # We use a map as a set
+
+# Check to see if the credentials directory exists, and make it if not.
+# We don't count this as a problem; if the user does not have the directory in
+# place, there are bigger problems with their directory structures anyway.
+if [[ ! -d "$CREDS" ]]; then
+    mkdir "$CREDS"
+fi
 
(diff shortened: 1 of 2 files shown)


scripts/setup_credentials.sh
```diff
@@ -0,0 +1,59 @@
...
+if ! grep -q app_id "$CREDS/github.toml"; then
+    echo 'app_id = -1' >> "$CREDS/github.toml"
+    ((problems++))
+fi
+if ! grep -q organization "$CREDS/github.toml"; then
+    echo 'organization = "YOUR_ORGANIZATION"' >> "$CREDS/github.toml"
+    ((problems++))
+fi
+if ! grep -q webhook_secret "$CREDS/github.toml"; then
+    echo 'webhook_secret = "NANI DA HECK"' >> "$CREDS/github.toml"
```
omar2535: lol 🤣

cheukyin699:
@omar2535 I opted to append to it because sometimes new variables will be added (and maybe removed), and so I would want it to be as non-destructive as possible. I'd wanna use it to update my existing credentials.

Of course, then, we'd want to have a better system for updating credentials and whatnot. Still thinking about the best way to solve that.

Good point on the "which files changed" part.

Issue #299, opened 2019-05-14
Team Commands empty database message


jacqueschen1:
Currently team commands don't return anything if the database doesn't contain any teams when calling `view` and other such commands. Will need to add return messages for these edge cases

PR #303, opened 2019-05-29
Added edge case for team list command

jacqueschen1:
# Pull Request

## Description

Added an edge case for team list command to return error messages if query returns no teams

## Testing

## Ticket(s)

Closes: #299 

### DIFF
--- a/tests/command/commands/team_test.py
+++ b/tests/command/commands/team_test.py
@@ -60,6 +60,12 @@ class TestTeamCommand(TestCase):
         self.assertTupleEqual(self.testcommand.handle("team list", user),
                               (self.testcommand.lookup_error, 200))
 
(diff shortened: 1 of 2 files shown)


tests/command/commands/team_test.py
```diff
@@ -60,6 +60,12 @@ def test_handle_list_lookup_error(self):
         self.assertTupleEqual(self.testcommand.handle("team list", user),
                               (self.testcommand.lookup_error, 200))
 
+    def test_handle_list_no_teams(self):
+        """Test team command list with no teams found."""
+        self.db.query.return_value = None
```
cheukyin699: Return value shouldn't be `None`. It should be `[]` empty list.

PR #309, opened 2019-06-01
Secured Slack endpoints

### LINKED ISSUE #295: Security issue with slash commands endpoints
Currently, we do not authenticate any of our slash command requests on the server-side. This means that any yahoo could go in (after acquiring a team-lead/admin's slack id), and just pretend to be a user and screw up all of our data and database.

This is extremely dangerous because, well, it can potentially screw up everything (good thing we have yet to use this in actual production). This is an issue of request validation, which must be fixed as soon as possible.

To the person assigned to this issue, [slack has a way of signing their requests][way]. They use the signing secret and combine it with the request body to create a unique hash. You only need to verify that it is indeed correct on the server before you process the request.

[way]: https://api.slack.com/docs/verifying-requests-from-slack

### LINKED ISSUE #307: Remove handle_app_mention endpoint
**Describe the bug**
The `handle_app_mention()` endpoint still exists in `server.py` from before we started using slash commands, even thought `handle_app_mention()` isn't even an existing function of `core` right now.

**To Reproduce**
It's just there.

**Expected behavior**
It shouldn't be there.

**Screenshots**

**Additional context**

rmcreyes:
# Pull Request

## Description

Used the `SlackEventAdapter`'s `verify_signature()` function to secure endpoints for slash commands and subscribed events from the Events API

## Testing

## Ticket(s)

Closes #295, Closes #307 

### DIFF
--- a/server/server.py
+++ b/server/server.py
@@ -90,13 +90,6 @@ def handle_github_webhook():
     return msg
 
 
-@slack_events_adapter.on("app_mention")
-def handle_app_mention(event):
-    """Handle a mention to @rocket."""
-    logging.info("Handled 'app_mention' event")
-    core.handle_app_mention(event)
-
-
 @slack_events_adapter.on("team_join")
 def handle_team_join(event):
     """Handle instances when user joins the Launchpad slack workspace."""
--- a/server/server.py
+++ b/server/server.py
@@ -73,9 +73,19 @@ def check():
 @app.route('/slack/commands', methods=['POST'])
 def handle_commands():
     """Handle rocket slash commands."""
-    txt = request.form['text']
-    uid = request.form['user_id']
-    return core.handle_app_command(txt, uid)
+    logging.info("Slash command received")
+    timestamp = request.headers.get("X-Slack-Request-Timestamp")
+    slack_signature = request.headers.get("X-Slack-Signature")
+    verified = slack_events_adapter.server.verify_signature(
+        timestamp, slack_signature)
+    if verified:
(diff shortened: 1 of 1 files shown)


server/server.py
```diff
@@ -73,9 +73,19 @@ def check():
...
     """Handle rocket slash commands."""
-    txt = request.form['text']
-    uid = request.form['user_id']
-    return core.handle_app_command(txt, uid)
+    logging.info("Slash command received")
+    timestamp = request.headers.get("X-Slack-Request-Timestamp")
+    slack_signature = request.headers.get("X-Slack-Signature")
+    verified = slack_events_adapter.server.verify_signature(
+        timestamp, slack_signature)
+    if verified:
```
cheukyin699: Please abstract this out into a different function, preferably `verify_slack_message(req)`

server/server.py
```diff
@@ -90,15 +100,16 @@ def handle_github_webhook():
...
 @slack_events_adapter.on("team_join")
 def handle_team_join(event):
     """Handle instances when user joins the Launchpad slack workspace."""
     logging.info("Handled 'team_join' event")
-    core.handle_team_join(event)
+    timestamp = request.headers.get("X-Slack-Request-Timestamp")
+    slack_signature = request.headers.get("X-Slack-Signature")
+    verified = slack_events_adapter.server.verify_signature(
+        timestamp, slack_signature)
+    if verified:
```
cheukyin699: ditto

PR #335, opened 2019-07-20
Karma command

omar2535:
# Pull Request

## Description

The long overdue karma command PR is here. 
Changes are 2 more commands. One for mentioning somebody using `/rocket @user` and another using `/rocket karma blah`. 
Commands implemented are
`rocket karma reset --all`
`rocket karma set @user 13`
`rocket karma view @user`
`rocket @user ++`
I opted to leave out reset for single users because set effectively does the same thing.
The docs have also been updated.

## Testing

N/A

## Ticket(s)

Affects #201


app/controller/command/commands/karma.py
```diff
@@ -0,0 +1,154 @@
...
+                return f"set {user.name}'s karma to {amount}", 200
+            else:
+                return self.permission_error, 200
+        except LookupError:
+            return self.lookup_error, 200
+
+    def reset_helper(self,
+                     user_id: str,
+                     reset_all: bool) -> ResponseTuple:
+        """Reset all user's karma."""
```
cheukyin699:
```suggestion
        """Reset all users' karma."""
```

app/controller/command/parser.py
```diff
@@ -1,80 +1,86 @@
...
+        # smart punctuation messing with argparse.
+        cmd_txt = ''.join(map(util.regularize_char, cmd_txt))
+        cmd_txt = util.escaped_id_to_id(cmd_txt)
+        s = cmd_txt.split(' ', 1)
+        if s[0] == "help" or s[0] is None:
+            logging.info("Help command was called")
+            return self.get_help(), 200
+        if s[0] in self.__commands:
+            return self.__commands[s[0]].handle(cmd_txt, user)
+        elif re.match("^[UW][A-Z0-9]{8}$", s[0]):
```
cheukyin699: You should probably put this into `utils/slack_parse.py`

app/controller/command/commands/mention.py
```diff
@@ -0,0 +1,44 @@
...
+        logging.info("Starting Mention command initializer")
+        self.parser = argparse.ArgumentParser(prog="Mention")
+        self.parser.add_argument("Mention")
+        self.facade = db_facade
+
+    def handle(self, command: str, user_id: str) -> ResponseTuple:
+        """Handle command by splitting into substrings."""
+        logging.debug('Handling Mention Command')
+        command_arg = shlex.split(command)
+        if(len(command_arg) <= 1):
```
cheukyin699:
```suggestion
        if len(command_arg) <= 1:
```

Too much java/c++ is toxic, I know.

app/controller/command/commands/karma.py
```diff
@@ -0,0 +1,154 @@
...
+        try:
+            user_command = self.facade.retrieve(User, user_id)
```
cheukyin699:
```suggestion
            user = self.facade.retrieve(User, user_id)
```

app/controller/command/commands/karma.py
```diff
@@ -0,0 +1,154 @@
...
+            if user_command.permissions_level == Permissions.admin:
```
cheukyin699: I don't like too many indents. Invert these if statements so that you return immediately, and proceed as normal afterwards.

Turn:

```python
def foo(a: bool, b: bool) -> str:
    if a:
        if b:
            return "BEE"
        else:
            return "CEE"
    return "ERROR"
```

Into:

```python
def foo(a: bool, b: bool) -> str:
    if not a:
        return "ERROR"

    if b:
        return "BEE"
    else:
        return "CEE"

```

app/controller/command/commands/mention.py
```diff
@@ -0,0 +1,44 @@
...
+            return self.help, 200
+        elif(command_arg[1] == "help"):
```
cheukyin699: So you are saying that `/rocket @cheukyin699 help` is possible? Should it be possible?

app/controller/command/commands/mention.py
```diff
@@ -0,0 +1,44 @@
...
+        elif(command_arg[1] == '++'):
+            return self.karma_mention_helper(user_id, command_arg[0])
+        else:
+            return self.help, 200
+
+    def karma_mention_helper(self,
+                             giver_id: str,
+                             receiver_id: str) -> ResponseTuple:
+        """Use the karma command to help add karma to a user."""
+        return KarmaCommand(self.facade).add_karma(giver_id, receiver_id)
```
cheukyin699: Why don't we move it here instead of creating a new object (and destroying it) every time?

app/controller/command/parser.py
```diff
@@ -1,80 +1,86 @@
...
+        self.__commands: Dict[str, UnionCommands] = {}
+        self.__facade = db_facade
+        self.__bot = bot
+        self.__github = gh_interface
+        self.__commands["user"] = UserCommand(self.__facade, self.__github)
+        self.__commands["team"] = TeamCommand(self.__facade,
+                                              self.__github, self.__bot)
+        self.__commands["token"] = TokenCommand(self.__facade, token_config)
+        self.__commands["karma"] = KarmaCommand(self.__facade)
+        self.__commands["mention"] = MentionCommand(self.__facade)
```
cheukyin699: It may be better to add `/rocket @user ++` documentation in `KarmaCommand` rather than view it with `/rocket mention help`.

app/controller/command/commands/mention.py
```diff
@@ -0,0 +1,55 @@
...
+    def add_karma(self, giver_id, receiver_id) -> ResponseTuple:
+        """Give karma from giver_id to receiver_id."""
+        logging.info("giving karma to " + receiver_id)
+        if giver_id == receiver_id:
+            return "cannot give karma to self", 200
+        try:
+            user = self.facade.retrieve(User, receiver_id)
+            user.karma += self.karma_add_amount
+            self.facade.store(user)
+            return f"gave 1 karma to {user.name}", 200
```
cheukyin699:
```suggestion
            return f"gave {self.karma_add_amount} karma to {user.name}", 200
```

I like how the noun `karma` is uncountable, so that we don't have to deal with tacking the `s` on it.

tests/app/controller/command/commands/mention_test.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertEqual(self.testcommand.handle('UFJ42EU67 --', user),
+                         (self.testcommand.unsupported_error, 200))
+
+    def test_handle_add_karma_to_another_user(self):
+        """Test handle command with karma to another user."""
+        user = User('U123456789')
+        user.name = 'U123456789'
+        self.mock_facade.retrieve.return_value = user
+        self.assertEqual(self.testcommand.handle(('UFJ42EU67 ++'), user),
+                         ("gave 1 karma to U123456789", 200))
```
cheukyin699: So the user `U123456789` ran the command `/rocket UFJ42EU67 ++`, and the result is that they gave karma to themselves??? Line 35 shouldn't return the user calling, it should return the user receiving. The handle command should have thrown a `TypeError` because it should handle the `user_id` and not the user object. Same with those above.

Also, there is no need for parentheses for the first argument of `.handle`.

app/controller/command/commands/mention.py
```diff
@@ -0,0 +1,55 @@
...
+        elif command_arg[1] == '++':
+            return self.karma_mention_helper(user_id, command_arg[0])
+        else:
+            return "unsupported usage", 200
+
+    def karma_mention_helper(self,
+                             giver_id: str,
+                             receiver_id: str) -> ResponseTuple:
+        """Use the karma command to help add karma to a user."""
+        return self.add_karma(giver_id, receiver_id)
```
cheukyin699: I meant that you could just remove `karma_mention_helper` and replace it with `add_karma`, since one is just passing the arguments off to the other.

app/controller/command/commands/mention.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def add_karma(self, giver_id, receiver_id) -> ResponseTuple:
```
cheukyin699:
```suggestion
    def add_karma(self, giver_id: str, receiver_id: str) -> ResponseTuple:
```

tests/app/controller/command/commands/mention_test.py
omar2535: Ah yes. Good catch!

tests/app/controller/command/commands/mention_test.py
```diff
@@ -20,30 +20,32 @@ def setUp(self):
...
-        self.mock_facade.retrieve.return_value = user
-        self.assertEqual(self.testcommand.handle(('UFJ42EU67 ++'), user),
+        reciever = User('U123456789')
+        reciever.name = 'U123456789'
+        giver = 'UFJ42EU67'
+        self.mock_facade.retrieve.return_value = reciever
+        self.assertEqual(self.testcommand.handle("U123456789 ++", giver),
                          ("gave 1 karma to U123456789", 200))
-        self.mock_facade.retrieve.assert_called_once_with(User, "UFJ42EU67")
+        self.mock_facade.retrieve.assert_called_once_with(User, 'U123456789')
```
cheukyin699: Since you return the receiver object, you can make an assertion about the receiver's karma directly. Something like:

```python
self.assertEqual(receiver.karma, 2)
```

PR #338, opened 2019-07-27
Refactor to use configuration from a singular source (environmental variables) instead of different toml files

### LINKED ISSUE #317: Confusion with credentials system

Credentials system is a bit convoluted and complicated and hard to understand; design around this.

I am very confused about our credentials system. We are encouraged (forced) to put our credentials into the `credentials/` directory, but most of the credentials remain in `config.toml`, and they overlap with the names recommended in the `credentials/` directory. Furthermore, we should be using `config.Credentials` a bit more to initialize things with their variables.

The configs and credentials.

Discussion on how these things should change in the comments below.

cheukyin699:
# Pull Request

## Description

We use a singular `.env` file (just like the olden days) that feeds all the environmental configuration things into our app through a proxy `Config` object. That way, we don't need 2 different objects for both configurations and credentials. We have a singular object, a singular file to change, and less confusion in general

## Testing

Tested with unit tests and running the server for a bit and seeing if I can break it. I couldn't really.

## Ticket(s)

Closes #317


docs/LocalDevelopmentGuide.md
```diff
@@ -61,54 +61,25 @@ policy. Finally, copy the provided access key ID and secret access key after
 creating the new user.
 
 Note: if you are in the `brussel-sprouts` Github team, you should already have
-AWS credentials.
+AWS credentials. Just ask.
```
rmcreyes: Nice

sample-env
```diff
@@ -0,0 +1,19 @@
+TESTING='True'
```
rmcreyes: Not necessary against this, but do we really want all of our configurations to go in one file? Probably makes things in the long run to be fair.

cheukyin699: I don't think we need to address that as of this moment. I think that using a single config file (or environment) should be good enough, since we don't have a multi-hundred line configuration files.

(The fact that we can still fit all of the variables on screen at the same time should say something about the amount of it; looking at you, Amazon).

rmcreyes: Aha fair enough

PR #344, opened 2019-08-15
Add an abstract base class for data models

### LINKED ISSUE #340: Abstract base class for models

Models should always have certain functions.

Right now, all our models have a specific set of functions, but the next model could easily not have that function. We need a way to ensure that they have the following functions:

- get_attachment(self) -> Dict[str, Any]
- static to_dict(obj) -> Dict[str, Any]
- static from_dict(Dict[str, Any]) -> obj
- is_valid(self) -> bool

The static functions along with the `is_valid` function are used specifically for the database. `get_attachment` is used for slack attachments.

rmcreyes:
# Pull Request

## Description

Added an ABC for data models with the abstract methods mentioned in #340 

## Testing

## Ticket(s)

Closes #340

### DIFF
--- a/app/model/base.py
+++ b/app/model/base.py
@@ -13,20 +13,6 @@ class RocketModel(ABC):
 
     @staticmethod
     @abstractmethod
-    def to_dict(model: 'RocketModel') -> Dict[str, Any]:
-        """
-        Convert data model object to dict object.
-
-        The difference with the in-built ``self.__dict__`` is that this is more
-        compatible with storing into NoSQL databases like DynamoDB.
-
-        :param model: the data model object
-        :return: the dictionary representing the data model
-        """
-        pass
-
-    @staticmethod
-    @abstractmethod
     def from_dict(d: Dict[str, Any]) -> 'RocketModel':
         """
         Convert dict response object to data model object.
@@ -36,13 +22,4 @@ class RocketModel(ABC):
         """
         pass
 
-    @staticmethod
-    @abstractmethod
-    def is_valid(model: 'RocketModel') -> bool:
-        """
-        Return true if this data model has no missing required fields.
-
-        :param model: data model object to check
-        :return: return true if this data model has no missing required fields
-        """
-        pass
+    # TODO: Add to_dict and and is_valid abstract methods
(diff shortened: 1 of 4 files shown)


app/controller/command/commands/__init__.py
```diff
@@ -3,5 +3,6 @@
 from app.controller.command.commands.team import TeamCommand
 from app.controller.command.commands.user import UserCommand
 from app.controller.command.commands.token import TokenCommand
+from app.controller.command.commands.project import ProjectCommand
 
-UnionCommands = Union[TeamCommand, UserCommand, TokenCommand]
+UnionCommands = Union[TeamCommand, UserCommand, TokenCommand, ProjectCommand]
```
cheukyin699: Should have no need for a union. We can use the parent class instead.

app/model/base.py
```diff
@@ -0,0 +1,48 @@
...
+    """Define the properties and methods needed for a data model."""
+
+    @abstractmethod
+    def get_attachment(self) -> Dict[str, Any]:
+        """Return slack-formatted attachment (dictionary) for data model."""
+        pass
+
+    @staticmethod
+    @abstractmethod
+    def to_dict(model: 'Model') -> Dict[str, Any]:
```
cheukyin699:
```suggestion
    def to_dict(model: 'RocketModel') -> Dict[str, Any]:
```

For every instance of `Model`.

app/controller/command/commands/__init__.py
rmcreyes: Should this be addressed in another ticket? This change is only to address models.

cheukyin699: sure. a smaller ticket. very welcome

cheukyin699:
This is weird. I'm going to look into this to see if there is a fix for it.

cheukyin699:
image

I recommend using non-static methods (another PR) and bounded types. See [this stackoverflow][so] for more details. Be sure to make another issue for refactoring some of the methods into non-static methods.

[so]: https://stackoverflow.com/questions/44640479/mypy-annotation-for-classmethod-returning-instance

rmcreyes:
Should those methods just be removed from the ABC for now then?

cheukyin699:
Yeah let's remove them for now and add them later; it's a good first step.

PR #346, opened 2019-08-18
Use Github Actions CI instead

### LINKED ISSUE #345: Use Github Actions CI

Since travis is no longer available, we should try and use Github Actions Python CI instead.

Just use it. Make sure it works.

Travis CI contacted us saying that our `ubclaunchpad` account was suspected of mining cryptocurrency using their services (as far as I know, we don't), and so they have suspended our account. This means all of Travis doesn't work, which means we no longer have a reliable way of checking pull requests. We need to replace it.

cheukyin699:
# Pull Request

## Description

**Give a brief description of your changes:** Use Github Actions CI instead of travis. Also makes tests slightly more stable (I ran `while pipenv run pytest -m db; do :; done` for about 10 minutes and it was still good, so it might be stable now, but we can always restart runs).

## Ticket(s)

Closes #345


.github/workflows/pythonpackage.yml
```diff
@@ -0,0 +1,36 @@
...
+        sudo apt-get update
+        sudo apt-get install -y default-jdk nmap rubygems
+        python -m pip install --upgrade pip
+        pip install pipenv
+        pipenv install --dev
+        sudo gem install mdl
+    - name: Start local DynamoDB
+      run: |
+        scripts/download_dynamodb_and_run.sh
+    - name: Test everything
```
rmcreyes: nice

PR #356, opened 2019-09-21
Replace nmap with a script

### LINKED ISSUE #355: Replace nmap as dependency

Replace `nmap` call with a simple python script.

After the whole rocket2 + travis debacle, we have come to realize that using `nmap` may lead to false positives in automated systems detecting fraudulent uses every time we check if a certain port is open. Therefore, we should remove `nmap` completely, and replace it with a simple python script to check things.

I would like the python script to take a port number as argument, and exit with `0` if that port is in use, and exit with any other exit code if that port is not in use.

Should just be as simple as removing the `nmap` install and replacing the `nmap | grep` with a python script call.

cheukyin699:
# Pull Request

## Description

**Give a brief description of your changes:** We replace usage of `nmap` with a python script to install fewer things. This might shave a few seconds off the build time.

## Testing

**If testing this change requires extra setup, please document it here:** Try running `scripts/build_check.sh` twice, once with local dynamodb and once without.

## Ticket(s)

Closes #355

### DIFF
--- a/scripts/build_check.sh
+++ b/scripts/build_check.sh
@@ -7,11 +7,10 @@ pipenv run mypy .
 mdl .
 
 # We use nmap to check if dynamodb is running locally
-# TODO: later should combine this with environmental variable checking (to see
-# if we can use a remote server instead).
-# XXX: Find a better way to check if dynamodb is running locally
 COV_OPTIONS="--mypy --cov=./ --cov-branch --cov-config .coverageac"
-if nmap localhost | egrep "8000.*http-alt"; then
+PORT_BUSY="pipenv run python scripts/port_busy.py 8000"
+if ${PORT_BUSY}; then
+    printf "DynamoDB detected. Running all tests.\n"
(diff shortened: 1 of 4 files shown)


scripts/build_check.sh
```diff
@@ -6,12 +6,11 @@ pipenv run pydocstyle .
...
 mdl .
 
-# We use nmap to check if dynamodb is running locally
-# TODO: later should combine this with environmental variable checking (to see
-# if we can use a remote server instead).
-# XXX: Find a better way to check if dynamodb is running locally
+# We use a script to check if dynamodb is running locally
 COV_OPTIONS="--mypy --cov=./ --cov-branch --cov-config .coverageac"
-if nmap localhost | egrep "8000.*http-alt"; then
+PORT_BUSY="pipenv run python scripts/port_busy.py 8000"
```
rwblickhan: Might be worthwhile to make this relative to the repo root dir (i.e. with an env var) in case people don't run it from there.

Issue #368, opened 2019-10-09
Properly separate responsibilities of team related commands and webhooks


rmcreyes:
**Please give a one-sentence summary of the cleanup you would like done.**
Both the team commands as well as the team and member seem to be trying to write data that should only be written once.

**Please give as many details as possible about the cleanup or refactoring.**
The intention is that when applicable, team commands would use the GitHub API to properly do action on the GitHub organization, allowing a webhook to come back down to handle the accesses to rocket's database. Remove writes to rocket's database in team commands and (where missing) add writes to rocket's database in the associated webhooks.

**Please list any additional context; in particular, list what areas of the code base this would affect.**
```
app\controller\command\commands\team.py
app\controller\webhook\github\events\team.py
app\controller\webhook\github\events\membership.py
```

cheukyin699:
To be clear, `/rocket team create` would both write to the database from the command and send a webhook that also writes to the database.

rmcreyes:
Seems like the only issue is with `membership.py`, since it fails to write to the database properly.

Issue #367, opened 2019-10-09
Incorrect use of facade retrieve


rmcreyes:
**Describe the bug**
The `delete_helper()` in the team commands attempts to call `retrieve()` the team name, when `query()` should be called instead.

PR #369, opened 2019-10-09
Massive bug fixing with rmcreyes

cheukyin699:
# Pull Request

## Description

**Give a brief description of your changes:** Touches on #367, #368. Tested on live database on slack all of `user` and `team` commands, organizational webhooks regarding users, team webhooks regarding users.

Lots of things went down. This was the result of a 5-hour-long call about ironing out bugs.

We also found out that Instagram now has dark mode. FYI.


cheukyin699:
Oh yes. A lot of things changed, even the APIs. The tests need to be rewritten. Or at least some of them.

PR #389, opened 2019-10-20
RandomChannelPromoter: use config.slack_notification_channel as the default, rename config.slack_announcement_channel

### LINKED ISSUE #386: Change scheduled "promote random channel" feature to use `#ot-random`
**Describe the bug**
The random channel promoter should use `#ot-random` instead of `#general`.

**Expected behavior**
To see the promoted channel in the `#ot-random` channel.

bobheadxi:
## Description

Use config.slack_bot_channel

## Testing

tests still pass (?)

## Ticket(s)

Fixes #386 


bobheadxi:
@cheukyin699 oh interesting - I had assumed that the configuration is for announcements, where `#ot-random` etc. might be appropriate. If it's a `#nt-` channel it would kind of defeat the purpose.

What kind of notifications are currently issued to `slack_bot_channel`? Should I add a new configuration option for `slack_announce_channel` and rename `slack_bot_channel` to `slack_notifications_channel`?

cheukyin699:
@bobheadxi The only thing it is currently used for is to announce when `rocket2` has restarted, though I personally cannot see any messages of that sort. I think adding a new configuration option and renaming the old one is good.

bobheadxi:
> though I personally cannot see any messages of that sort

could be a bug 🙃 

> I think adding a new configuration option and renaming the old one is good.

I'll make the change 🚀

sample-env
```diff
@@ -1,6 +1,6 @@
 TESTING='True'
 
-SLACK_BOT_CHANNEL='#rocket2'
+SLACK_NOTIFICATION_CHANNEL='#rocket2'
```
cheukyin699:
```suggestion
SLACK_NOTIFICATION_CHANNEL='#rocket2'
SLACK_ANNOUNCEMENT_CHANNEL='#rocket2-announcements'
```

tests/config/config_test.py
```diff
@@ -63,7 +63,7 @@ def test_incomplete_config(incomplete_config):
         Config()
         assert False
     except MissingConfigError as e:
-        missing_fields = ['SLACK_BOT_CHANNEL', 'SLACK_SIGNING_SECRET',
+        missing_fields = ['SLACK_NOTIFICATION_CHANNEL', 'SLACK_SIGNING_SECRET',
                           'SLACK_API_TOKEN']
```
cheukyin699:
```suggestion
                          'SLACK_API_TOKEN', 'SLACK_ANNOUNCEMENT_CHANNEL']
```

bobheadxi:
@cheukyin699 will this break the deployment (renaming `config.slack_bot_channel` to `config.slack_notification_channel`)?

cheukyin699:
@bobheadxi no. it would break the alpha build tho.

PR #399, opened 2019-10-29
Fix for iOS converting double-hyphens to dashes

### LINKED ISSUE #384: Double-dash not recognized by parser in iOS (Unicode)
**Describe the bug**
iOS would merge the double-dash into a single longer-dash as part of its keyboard settings and this would lead to rocket not recognizing double-dashes in the commands.

**To Reproduce**
run command '/rocket user edit --name "name"' in an iOS Device

**Expected behavior**
For rocket to understand iOS' long-dash as a double-dash.

phillytan:
# Pull Request

## Description

Added a helper function in `slack_parser.py` which replaces all instances of iOS dashes with the double-hyphen since iOS converts double-hyphens to dashes as shown here: https://support.apple.com/kb/PH26552?viewlocale=en_US&locale=en_US.

## Testing

None

## Ticket(s)

Fixes #384

### DIFF
--- a/app/controller/command/parser.py
+++ b/app/controller/command/parser.py
@@ -53,6 +53,7 @@ class CommandParser:
         # smart punctuation messing with argparse.
         cmd_txt = ''.join(map(util.regularize_char, cmd_txt))
         cmd_txt = util.escaped_id_to_id(cmd_txt)
+        cmd_txt = util.ios_dash(cmd_txt)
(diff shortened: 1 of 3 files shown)


cheukyin699:
Hmm. You may want to add some unit tests to this. Like what would happen if you have more than one of those unicode dashes.

Issue #222, opened 2019-02-13
Refactor business logic to accommodate HTTP API


rwblickhan:
Our business logic (as expressed in modules like `UserCommand` etc) is tied pretty intimately to Slack messaging. Thus, if we introduce an HTTP API, we will need to duplicate a fair amount of functionality. We should refactor so that the business logic is independent of Slack messaging. Then the HTTP API can access the business logic directly while `UserCommand` and so on can access the business logic and format it appropriately for messaging to Slack.

rwblickhan:
![IMG_3990](https://user-images.githubusercontent.com/13651149/54872205-b71ea380-4d7d-11e9-87c3-d8fcacec594f.jpg)

Here's our notes on implementing it.

cheukyin699:
To Do:

- [ ] `UserCommands` #419
- [ ] `TeamCommands` #419
- [ ] `ProjectCommands`
- [ ] `KarmaCommands`

PR #419, opened 2020-01-13
Refactor business logic of user and team commands

rmcreyes:
# Pull Request

## Description

The purpose of this PR is to port over the existing command logic to a more general space to be used more easily by different interfaces (ie. Slack client, REST APIs, etc.).  This PR is only to provide the re implemented logic, as it may be easier to enable all this little by little in small PRs in the future. This PR also only covers user and team command logic.

## Testing

Additional tests were created.

## Ticket(s)

Affects #222

### DIFF
--- a/api/command/mixins/team.py
+++ b/api/command/mixins/team.py
@@ -68,7 +68,8 @@ class TeamCommandApis:
         :raises: LookupError if the calling user or tech lead cannot be found
                  in the database
         :raises: PermissionError if the calling user has insufficient
-                 permissions to create a team
+                 permissions to create a team, or if the specified user with
+                 lead_id does not have the permission to be a lead
         :raises: SlackAPIError if a channel name is provided by an error
                  is encountered retrieving the members of that channel
         :raises: GithubAPIException if an error occurs on team creation or
@@ -133,22 +134,26 @@ class TeamCommandApis:
                           f"added to {gh_team_name}")
 
         if lead_id is not None:
-            # XXX: Should check be done to ensure user has lead status?
             lead = self._db_facade.retrieve(User, lead_id)
 
-            lead_in_team = self._gh_interface.has_team_member(
-                lead.github_username,
-                gh_team_id)
-            if not lead_in_team:
-                self._gh_interface.add_team_member(lead.github_username,
-                                                   gh_team_id)
+            if check_permissions(lead, None):
+                lead_in_team = self._gh_interface.has_team_member(
+                    lead.github_username,
+                    gh_team_id)
+                if not lead_in_team:
+                    self._gh_interface.add_team_member(lead.github_username,
+                                                       gh_team_id)
 
-            # XXX: Should lead also be added as member via `add_member`?
-            team.add_team_lead(lead.github_id)
-            logging.debug(f"User with ID {lead_id} set as tech lead of "
-                          f"{gh_team_name}")
-        # XXX: In the case where a channel name is provided, this may result
-        #      in setting a tech lead who is not a part of the team
+                # XXX: Should lead also be added as member via `add_member`?
+                team.add_team_lead(lead.github_id)
+                logging.debug(f"User with ID {lead_id} set as tech lead of "
+                              f"{gh_team_name}")
+            else:
+                msg = f"User specified with lead ID {lead_id} has S" \
+                    f" permission level {str(lead.permissions_level)}, " \
+                    "insufficient to lead a team!"
+                logging.error(msg)
+                raise PermissionError(msg)
         else:
             team.add_team_lead(command_user.github_id)
             logging.debug(f"Calling user with ID {command_user.github_id} set"
--- a/api/command/mixins/team.py
+++ b/api/command/mixins/team.py
@@ -144,7 +144,7 @@ class TeamCommandApis:
                     self._gh_interface.add_team_member(lead.github_username,
                                                        gh_team_id)
 
-                # XXX: Should lead also be added as member via `add_member`?
+                team.add_member(lead.github_id)
                 team.add_team_lead(lead.github_id)
                 logging.debug(f"User with ID {lead_id} set as tech lead of "
                               f"{gh_team_name}")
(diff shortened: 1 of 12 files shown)


api/command/mixins/team.py
```diff
@@ -0,0 +1,701 @@
...
+        if added:
+            logging.info(f"Member successfully added: {team.__str__()}")
+        else:
+            logging.error("Member add unsuccessful")
+        return added
+
+    def team_remove(self,
+                    user_id: str,
+                    gh_team_name: str,
+                    slack_id: str) -> bool:
```
cheukyin699: Same here with unclear `user_id` and `slack_id`

Maybe have it called caller_id (caller is implicitly a user) or something.

api/command/mixins/team.py
```diff
@@ -0,0 +1,701 @@
...
+
+
+class TeamCommandApis:
+    """Encapsulate the various APIs for team command logic."""
+
+    def __init__(self):
+        """Declare the interfaces needed."""
+        self._db_facade = None
+        self._gh_interface = None
+        self._slack_client = None
```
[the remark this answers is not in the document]
cheukyin699: Also we need to provide types for these interfaces so that `mypy` knows function return types (so that you don't have to `cast` everything as much).

api/command/mixins/team.py
```diff
@@ -0,0 +1,701 @@
...
+        if display_name is not None:
+            logging.debug(f"Attaching display name {display_name} "
+                          f"to {gh_team_name}")
+            team.display_name = display_name
+
+        if platform is not None:
+            logging.debug(f"Attaching platform {platform} to {gh_team_name}")
+            team.platform = platform
+
+        edited = cast(bool, self._db_facade.store(team))
```
cheukyin699: See the first comment (I think its the first), which talks about specifying the type so that we don't have to cast.

api/command/mixins/team.py
```diff
@@ -0,0 +1,701 @@
...
+            raise GithubAPIException(msg)
+        local_team_dict = dict((team.github_team_id, team)
+                               for team in local_teams)
+        remote_team_dict = dict((team.github_team_id, team)
+                                for team in remote_teams)
+
+        # remove teams not in github anymore
+        for local_id in local_team_dict:
+            if local_id not in remote_team_dict:
+                # XXX: Should this be a local delete instead of a Github delete?
```
cheukyin699: Yes this should be a local delete.

api/command/mixins/team.py
```diff
@@ -0,0 +1,503 @@
...
+                    member = self._db_facade.retrieve(User, member_id)
+                    self._gh_interface.add_team_member(member.github_username,
+                                                       gh_team_id)
+                    team.add_member(member.github_id)
+                    logging.debug(f"Member with ID {member.slack_id} added "
+                                  f"to {gh_team_name}")
+                except LookupError:
+                    logging.warning(f"Member with ID {member_id} not found, "
+                                    "ignoring...")
+                    pass
```
cheukyin699: Probably a good idea to use a [bulk retrieve](https://rocket2.readthedocs.io/en/latest/docs/api/db.html#db.facade.DBFacade.bulk_retrieve) and check later if the lengths match.

api/command/mixins/team.py
```diff
@@ -0,0 +1,503 @@
...
+                if not lead_in_team:
+                    self._gh_interface.add_team_member(lead.github_username,
+                                                       gh_team_id)
+
+                team.add_member(lead.github_id)
+                team.add_team_lead(lead.github_id)
+                logging.debug(f"User with ID {lead_id} set as tech lead of "
+                              f"{gh_team_name}")
+            else:
+                msg = f"User specified with lead ID {lead_id} has S" \
```
cheukyin699:
```suggestion
                msg = f"User specified with lead ID {lead_id} has" \

```

What's the `S` for?

api/command/mixins/team.py
```diff
@@ -0,0 +1,503 @@
...
+        else:
+            team.add_team_lead(command_user.github_id)
+            logging.debug(f"Calling user with ID {command_user.github_id} set"
+                          f" as tech lead of {gh_team_name}")
+
+        created = self._db_facade.store(team)
+        if created:
+            logging.info(f"Team succesfully created: {team.__str__()}")
+        else:
+            logging.error("Team creation unsuccessful")
```
cheukyin699: I guess, a bit of a misnomer, because this only checks if the team is stored into DDB, but doesn't check if it has been created on the Github side. Maybe rename to `Team successfully stored`.

cheukyin699: Actually, the database logs if the store is successful or not. No need to log it here.

api/command/mixins/team.py
```diff
@@ -0,0 +1,503 @@
...
+
+        self._gh_interface.add_team_member(add_user.github_username,
+                                           team.github_team_id)
+        team.add_member(add_user.github_id)
+
+        added = self._db_facade.store(team)
+        if added:
+            logging.info(f"Member successfully added: {team.__str__()}")
+        else:
+            logging.error("Member add unsuccessful")
```
cheukyin699: Same here

api/command/mixins/team.py
rmcreyes: Nothing, removed

api/command/mixins/team.py
rmcreyes: Removed

api/command/mixins/team.py
rmcreyes: This and others removed

api/command/mixins/team.py
rmcreyes: What would be bulk retrieved (and with what parameters)?

cheukyin699:
```python
for member in self._db_facade.bulk_retrieve(channel_member_ids):
	pass
```

rmcreyes: Tried implementing what I thought you might want, let me know if there's anything to change in a follow up change request

PR #461, opened 2020-05-06
Have team refresh command add all members to all

### LINKED ISSUE #459: add all members to @ubclaunchpad/all
Old Rocket maintains a team, https://github.com/orgs/ubclaunchpad/teams/all, that makes it easy to "inner source" repositories (make repos visible to everyone within Launch Pad). [code](https://sourcegraph.com/github.com/ubclaunchpad/rocket@0eb8ecf425c01cd20672c549bde19c6bab0ec020/-/blob/plugins/core/set.go#L92)

We need Rocket2 to add everyone to this team as well

This is causing issues with folks access private repos - see https://ubclaunchpad.slack.com/archives/C061Q84SW/p1588451606003600

### LINKED ISSUE #460: Team Refresh removes non-existent team from Github instead of from local database
**Describe the bug**

When we execute `/rocket team refresh`, if there is an inconsistency, it doesn't remove the team from our local database, but rather the (non-existent) team is removed from Github. Which probably causes an error and halts the entire process due to the entire thing being enclosed in a try-catch block.

All speculation of course. But since we use Github as the sole source of truth, there is no way that in the process of updating our databases, that we have to modify anything on Github's end.

cheukyin699:
# Pull Request

## Description

**Give a brief description of your changes:** We want the refresh command to add all members to the team (that old rocket maintained) `all`.

We also fix a bug where we remove a nonexistent team from Github if we see that that team exists in our database but not in Github. We now do the opposite, and remove the team from our local database.

## Ticket(s)

Closes #459 
Closes #460

### DIFF
--- a/app/controller/command/commands/team.py
+++ b/app/controller/command/commands/team.py
@@ -574,7 +574,7 @@ class TeamCommand(Command):
             # remove teams not in github anymore
             for local_id in local_team_dict:
                 if local_id not in remote_team_dict:
-                    self.gh.org_delete_team(local_id)
+                    self.facade.delete(Team, local_id)
                     num_deleted += 1
                     modified.append(local_team_dict[local_id].get_attachment())
 
--- a/app/controller/command/commands/team.py
+++ b/app/controller/command/commands/team.py
@@ -620,7 +620,8 @@ class TeamCommand(Command):
 
     def refresh_all_team(self):
         """
-        Refresh the 'all' team.
+        Refresh the 'all' team - this team is used to track all members.
+        See https://github.com/orgs/ubclaunchpad/teams/all
 
         Should only be called after the teams have all synced, or bugs will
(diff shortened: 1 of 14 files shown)


app/controller/command/commands/team.py
```diff
@@ -610,3 +617,36 @@ def refresh_helper(self, user_id) -> ResponseTuple:
             f"{num_deleted} deleted. Wonderful."
         ret = {'attachments': modified, 'text': status}
         return ret, 200
+
+    def refresh_all_team(self):
+        """
+        Refresh the 'all' team.
```
bobheadxi, line 623: For future reference perhaps?

```suggestion
        Refresh the 'all' team - this team is used to track all members of UBC Launch Pad.
        See https://github.com/orgs/ubclaunchpad/teams/all
```

PR #484, opened 2020-09-24
feat: google drive permissions sync

### LINKED ISSUE #284: Google drive permissions management
we should sync emails from teams onto access to folders

bobheadxi:
# Pull Request

## Description

Enables an optional step to `team refresh`, enabled by providing `GCP_SERVICE_ACCOUNT_CREDENTIALS`, to synchronize team folders by sharing each team's configured `folder` to all emails provided by that team's members.

The current _intended_ use case for this is _only_ for the top-level folders indicated in https://docs.ubclaunchpad.com/handbook/tools/drive - i.e I will be assigning the following folders:

* `all`: `Projects`
* `leads`: `UBC Launch Pad`
* `design`: `Design`
* `strategy`: `Strategy`

As such, no care has been taken for certain edge cases (ie "what if I *don't* want to unshare with anyone not on a rocket team?") - for this iteration of this feature, I am going to keep it simple.

The goal is to remove one of the last steps that leads have to do from onboarding - see https://github.com/ubclaunchpad/docs/pull/175 . It currently requires a refresh - see https://github.com/ubclaunchpad/rocket2/issues/490, https://github.com/ubclaunchpad/rocket2/issues/495

## Testing

Simple unit test provided. ~Working on a clicky-click test as well.~

Update - verified this works!

image

## Ticket(s)

Closes #284 


app/controller/command/commands/team.py
```diff
@@ -650,3 +669,30 @@ def refresh_all_team(self):
...
+            for m in t.members:
+                if len(m.email) > 0:
+                    emails.append(m.email)
+
+            if len(emails) > 0:
+                logging.info("Synchronizing permissions for "
+                             + f"{t.github_team_name}'s folder ({t.folder}) "
+                             + f"to {emails}")
+                self.gcp.set_drive_permissions(
+                    t.folder, t.github_team_name, emails)
```
cheukyin699, line 698: You may want to run a more specialized version of this function, specific to the team that called with a `--folder`, to sync permissions for that team's team members only.

Something similar to what happens when we do `/rocket user edit --github` and we do an invite for that github user to join our organization.

app/model/team.py
```diff
@@ -31,6 +32,7 @@ def get_attachment(self):
             ('Github Team Name', self.github_team_name),
             ('Display Name', self.display_name),
             ('Platform', self.platform),
+            ('Folder', self.folder),
```
cheukyin699, line 35: You will also need to modify the `Team.to_dict()` function so that we can properly store the folder.

chuck-sys (changes requested):
Needs a bit more testing.

app/model/team.py
bobheadxi: c5d65981dd8967ead7789563feea61b5c2159f28

app/controller/command/commands/team.py
bobheadxi: Do you think it's necessary at the moment? Happy to implement this in a follow-up instead :)

chuck-sys (commented):
Okay so here's the thing. `/rocket team refresh` takes around 20 seconds to execute (last I checked). The more members we have in Launch Pad, the longer it takes, `O(n)`. We'll still need ways for users to receive the permission sharing automatically, whether if it is a user just adding their emails, or a team adding a folder ID into the system.

I am very adamant about **not** using `/rocket team refresh` unless we really need to because ... there are lots of side effects that I don't know about, that may be disruptive. For example, today I found out that it would forcefully add a user back into the organization if the user was in ddb but was already out of the org (one possibility would be if the user ID was still in a team in ddb).

Will have to implement this as a follow-up. For now I'm approving this.

chuck-sys (approved):
Okay so here's the thing. `/rocket team refresh` takes around 20 seconds to execute (last I checked). The more members we have in Launch Pad, the longer it takes, `O(n)`. We'll still need ways for users to receive the permission sharing automatically, whether if it is a user just adding their emails, or a team adding a folder ID into the system.

I am very adamant about **not** using `/rocket team refresh` unless we really need to because ... there are lots of side effects that I don't know about, that may be disruptive. For example, today I found out that it would forcefully add a user back into the organization if the user was in ddb but was already out of the org (one possibility would be if the user ID was still in a team in ddb).

Will have to implement this as a follow-up. For now I'm approving this.

bobheadxi:
@cheukyin699 agreed and noted! I've created https://github.com/ubclaunchpad/rocket2/issues/490 to cover that

PR #492, opened 2020-09-25
make sure env vars are seen by docker compose

chuck-sys:


bobheadxi (approved):
thanks! linking for reference - these vars were added in https://github.com/ubclaunchpad/rocket2/pull/484

PR #499, opened 2020-09-26
gcp: attempt to exclude inherited permissions from deletion

### LINKED ISSUE #497: drive: permission invites get re-sent on occasion
I can't seem to fetch permissionDetails to determine if a permission is inherited - something about the difference between personal drives and team drives? https://developers.google.com/drive/api/v3/reference/permissions#resource

One solution might be to not delete permissions at all => https://github.com/ubclaunchpad/rocket2/issues/497#issuecomment-705993259

bobheadxi:
# Pull Request

## Description

Doesn't really work at time of PR opening - see #497 . This might be a Google Drive limitation - requires more investigation.

## Testing

## Ticket(s)

Closes #497 


bobheadxi:
we do not use shared drives, so this is an incorrect approach

PR #517, opened 2020-09-29
Add command execution timing metrics (along with infrastructure for future metrics)

### LINKED ISSUE #480: Custom timing metrics per command execution
Add ability to submit metrics of time it takes to execute current command.

Using AWS CloudWatch `put-metric-data` API endpoint, we can submit metrics. [Here's a link to the api](https://docs.aws.amazon.com/cli/latest/reference/cloudwatch/put-metric-data.html). Just do it in python, shouldn't be too difficult.

chuck-sys:
# Pull Request

## Description

**Give a brief description of your changes:** Add command execution timing metrics for better clarity. Here are some metrics collected while testing:

image

## Ticket(s)

Closes #480


interface/cloudwatch_metrics.py
```diff
@@ -0,0 +1,40 @@
+import logging
+import boto3
+from config import Config
+
+
+class CWMetrics:
+    def __init__(self, config: Config):
+        if config.aws_local:
+            self.cw = None
```
bobheadxi, line 9: 💪

app/controller/command/parser.py
```diff
@@ -59,27 +63,37 @@ def handle_app_command(self,
...
             logging.error("app command triggered incorrectly")
             v = self.get_help()
         if isinstance(v[0], str):
             response_data: Any = {'text': v[0]}
         else:
             response_data = v[0]
+
+        # Submit metrics
+        duration_taken_ms = time.time() * 1000 - start_time_ms
+        self.__metrics.submit_cmd_mstime(cmd_name, duration_taken_ms)
```
bobheadxi, line 95: For the command:

```
/rocket team view bob
```

Is `cmd_name` just `team`, or `view`, or `team view bob`? I feel like it should encapsulate the first two (`team view`)

cheukyin699: `cmd_name` should be `team` and `help` and `user`. with the way things are set up right now, it won't be easy to encapsulate the first two, because some commands (eg mention, help) don't even have that. tho I can add a catch-all for those.

bobheadxi: I think having the dimension for just `team` would not give us very useful metrics, since that covers such a broad range of functionality. For example, `team refresh` being terrible doesn't mean we need to look at all the team commands

> because some commands (eg mention, help) don't even have that

Ah good point - though we could just leave that empty right?

cheukyin699: > though we could just leave that empty right?

I... don't know about empty. Maybe make it "not specified".

bobheadxi: I mean if we join it as a single string:

```
/rocket team view bob -> "team view"
/rocket help -> "help "
/rocket user -> "user"
/rocket user view -> "user view"
```

cheukyin699: May be too generic. There can be `/rocket user -h` as well.

bobheadxi: Damn, maybe we can do a "dumb" filter to just not log help events? Ignore if `help` or `-h`

bobheadxi:
Would be good to document this in https://sourcegraph.com/github.com/ubclaunchpad/rocket2/-/blob/docs/LocalDevelopmentGuide.rst#L184-206 as well (though I feel like some of that should be moved into the Configuration section instead)

PR #526, opened 2020-10-01
build: make dependency resolution faster

chuck-sys:
# Pull Request

## Description

**Give a brief description of your changes:** Cache pipenv directory for (hopefully) faster build times. I am tired of builds taking 4 minutes because pipenv wants to install everything and pipenv is slow as heck.


cheukyin699:
I'm starting to look into this a bit more, and it seems like the only other option to speed up dependencies (even when we cache the entire pipenv cache directory) is to switch to poetry. Will be looking into that in this PR.

cheukyin699:
Turns out that `awscli` is really picky with how it does `boto3` dependencies. By removing it, it saves some time.

cheukyin699:
`poetry` lacks a few things that Rocket 2 uses. For example, `poetry` doesn't implicitly import `.env` files as the environment. They also deal with scripts differently.

PR #534, opened 2020-10-04
chore: rename main CI workflow to 'pipeline'

bobheadxi:
# Pull Request

## Description

It's starting to feel a bit awkward to _not_ be on the "no master" bandwagon (especially now that GitHub default branches for new repos default to `main`), so while I don't really care about changing the default branch... this feels a bit egregious:

image

## Testing

## Ticket(s)

Affects #


chuck-sys (approved):
ok master

PR #546, opened 2020-10-06
docs: add undoc'd fns, update cmd apis, stop recommending localtunnel

chuck-sys:
Add previously undocumented functions, update user command API with new options, and stop recommending localtunnel because it is very problematic.

https://rocket2.readthedocs.io/en/cheukyin699-update-docs/

### DIFF
--- a/docs/LocalDevelopmentGuide.rst
+++ b/docs/LocalDevelopmentGuide.rst
@@ -127,8 +127,9 @@ ngrok gives you. Note that it is very important to use the HTTPS URL,
 Localtunnel
 ~~~~~~~~~~~
 
-    Note: We no longer recommend localtunnel because of reliability issues.
-    Tread with caution.
+    Note: Localtunnel has reliability issues, which include crashing
+    Rocket 2 after issuing a single command sometimes (for no discernable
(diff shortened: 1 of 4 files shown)


docs/LocalDevelopmentGuide.rst
```diff
@@ -127,6 +127,9 @@ ngrok gives you. Note that it is very important to use the HTTPS URL,
 Localtunnel
 ~~~~~~~~~~~
 
+    Note: We no longer recommend localtunnel because of reliability issues.
+    Tread with caution.
```
bobheadxi, line 131: Super unfortunate that localtunnel doesn't perform well, but it sure beats changing my URLs when using ngrok imo - can we update the disclaimer to just say that localtunnel has reliability issues, instead of changing the recommendation? I would still recommend it for folks who do sporadic contributions, because of the ability to keep your URL

tests/memorydb.py
```diff
@@ -94,6 +94,17 @@ def get_key(m: T) -> str:
 
 
 class MemoryDB(DBFacade):
```
bobheadxi, line 96: beautiful hehe

image

PR #548, opened 2020-10-10
team: add drive sync events on add, remove, edit

bobheadxi:
Add drive sync events for `team add`, `team remove`, and `team edit --folder`.

## Ticket(s)

n/a

## Details

Folders are not shared as expected when a user is added to a team, leading to some confusing misinformation I gave to @sandyklc about Rocket being able to take care of things.


chuck-sys (approved):
I was about to talk about the unused variable on line 552, but then it went away. LGTM.

Issue #535, opened 2020-10-04
when to use slack mentions as arguments is common source of confusion


bobheadxi:
There are a few threads and questions raised where it seems unclear when to use Slack mentions (e.g. `@robert`), for example when providing `--lead` to `team create`

Maybe there's some programatic checks we can add, or some improved documentation, to alleviate this

cheukyin699:
I think that programmatic checks should be the way to go. Currently the state of error reporting from the app is pretty bad, with unhelpful errors that make it difficult to debug (some of them anyway). Planning on a small rewrite on where the errors get thrown and taken so that error messages are better separated and have more context (in both slack and logs). See #515.

Issue #203, opened 2019-02-08
Organization-agnosticity


bobheadxi:
Similar to a ticket I created over at old™️ Rocket (https://github.com/ubclaunchpad/rocket/issues/84), it might be cool make Rocket more org-agnostic. I don't think we're the only club of our kind out there, and I'm sure this kind of team/project-management Slack bot might find use amongst other organizations as well.

Things like not hard-coding `ubclaunchpad` and more robust configuration options might be enough to make this a thing.

This is *low-priority*, defs focus on our use case first, but it's something I think would be cool to say about Rocket 2 down the line 💯

rwblickhan:
Where is `ubclaunchpad` hardcoded?

bobheadxi:
> not hard-coding `ubclaunchpad` (which I see you guys already have)

wording is finicky - I mean you guys *already have not hardcoded*

maybe i need to take more ENGL courses?

rwblickhan:
Ah, haha, I see what you mean now. You mean "not hard-coding `ubclaunchpad` is a feature you guys already have", not "`not hard-coding `ubclaunchpad`, which you already did".

English is hard man 🤷‍♀️

PR #553, opened 2020-10-12
feat: Slack pairings

### LINKED ISSUE #514: donut-style random pairing for members in a certain channel
Donut is an app that pairs random folks together to talk. That's basically all it does, and costs $50 a month. Now, Rocket can already send messages, and it is aware of all our members and channels... 😉

tarikeshaq:
fixes #514 , closes https://github.com/ubclaunchpad/ideas/issues/8

## Details

Creates basic donut style matching for Rocket.

### What implemented so far:
- A donut scheduler that currently runs every minute (for debugging purposes, will bump to a reasonable number once this is ready to land)
- Some initial refactoring to allow the schedulers to have access to the db
- Slack api calls needed to: 
    - Get a slack channel's id by name
    - Create a private conversation given a list of users
- Basic persistence of pairings in a new pairings table
### TODOs before this is ready:
- [x] Add persistence, to prevent users from getting matched again
- [x] ~~Add TTL to entries in the Pairings table~~ I can file a follow up for this, we can live without it for now
- [x] Add tests
   - For slack API calls
   - For persistence
- [ ] Add documentation
   - General documentation for the feature
   - For the new environment variable
   - Documentation for the new pairings table

### Where I am currently:
~~Contemplating if I should add a table for the pairings or if it should be a part of the `User` model, leaning towards the latter with possibly opening a ticket for the former beyond this pr.
Went with creating a new table 😄 Now working on adding the TTL and making sure edge cases don't explode on us~~

I'll be updating the above as I go.


config/__init__.py
```diff
@@ -17,7 +17,7 @@ class Config:
         'SLACK_API_TOKEN': 'slack_api_token',
         'SLACK_NOTIFICATION_CHANNEL': 'slack_notification_channel',
         'SLACK_ANNOUNCEMENT_CHANNEL': 'slack_announcement_channel',
-
+        'SLACK_DONUT_CHANNEL': 'slack_donut_channel',
```
bobheadxi, line 20:
```suggestion
        'SLACK_PAIRINGS_CHANNEL': 'slack_pairings_channel',

```

etc - not that the chances of being sued is very high, but we should probably stick on the safe side with a more generic name like "pairings" for this feature. Also, not everyone is familiar with donut either

bobheadxi: (or `matchings`)

app/scheduler/modules/donut.py
```diff
@@ -0,0 +1,63 @@
...
+        self.bot = Bot(WebClient(config.slack_api_token),
+                       config.slack_notification_channel)
+        self.channel_id = self.bot.get_channel_id(config.slack_donut_channel)
+        self.facade = facade
+
+
+    def get_job_args(self) -> Dict[str, Any]:
+        """Get job configuration arguments for apscheduler."""
+        return {'trigger':      'cron',
+                'minute': '*',
```
cheukyin699, line 32:
```suggestion
                'minute':       '*',
```

Needs more spaces

cheukyin699:
> Contemplating if I should add a table for the pairings or if it should be a part of the `User` model, leaning towards the latter with possibly opening a ticket for the former beyond this pr.

IMO I think you should go with the former (though the difference in work is pretty significant). Seems less hacky and wouldn't leak too much information (example: `/rocket user view` can potentially see who you matched with).

And it would set the precedent for similar things in the future.

app/scheduler/modules/pairing.py
```diff
@@ -0,0 +1,63 @@
+"""Match two Launchpad member for a private conversation"""
```
bobheadxi, line 1:
```suggestion
"""Match two Launch Pad members for a private conversation"""
```

😢 https://docs.ubclaunchpad.com/handbook/design/branding#launch-pad

app/scheduler/modules/pairing.py
```diff
@@ -0,0 +1,63 @@
...
+        self.bot = Bot(WebClient(config.slack_api_token),
+                       config.slack_notification_channel)
+        self.channel_id = self.bot.get_channel_id(config.slack_pairing_channel)
+        self.facade = facade
+
+
+    def get_job_args(self) -> Dict[str, Any]:
+        """Get job configuration arguments for apscheduler."""
+        return {'trigger':      'cron',
+                'minute': '*',
```
bobheadxi, line 33: I'm guessing this is just temporary for testing, but could also just make this a configuration option! `SLACK_PAIRING_FREQUENCY` for example

interface/slack.py
```diff
@@ -110,6 +110,43 @@ def send_event_notif(self, message):
...
+        Retrieves a channel's id given it's name
+
+        :param channel_name: The name of the channel
+
+        :raise SlackAPIError if no channels were found with the name `channel_name`
+
+        :return the slack id of the channel
+        """
+        # We strip away the "#" in case it was provided with the channel name
+        channel_name = channel_name.replace("#", "")
```
bobheadxi, line 141: There are two cases I've seen where we ask for channel name as input:

* from Slack commands, in which case we should require the appropriate Slack mention (#535)
* from configuration, in which case we should enforce the `#` prefix when loading the configuration - there is at least one other config option where this happens, and we don't strip prefixes it seems (at least from a quick search: https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/ubclaunchpad/rocket2%24+replace%28%22%23%22%2C+%22%22%29+OR+replace%28%27%23%27%2C+%27%27%29&patternType=literal). We don't enforce this at config load time right now though, but I think that's a better option

tldr

```suggestion
```

bobheadxi (commented):
nice progress! 🎉 

(also, it seems your commit authors arent attached to your GitHub, not sure if intentional?)

interface/slack.py
tarikeshaq: Hmmm yea the reason I'm keeping it here is that I'm doing the matching on channel names on my side (as apposed to on the slack API's side) and since the call to slack's `conversations_list` returns without the `#` prefix I have to deal with it 😢 A different approach could be to force it to be without the `#` in the env variables

tarikeshaq:
OK I got some time to work on this again (after like 20 days 😢) and Added a few tests.

I think this is ready for a round of review so I'll change it from a draft. My next focus will be docs and addressing any feedback that pops up. No rush on the review, we are all busy and this is decently large.

app/model/pairing.py
```diff
@@ -0,0 +1,111 @@
...
+class Pairing(RocketModel):
+    """Represent a pairing and related fields and methods."""
+
+    def __init__(self,
+                 user1_slack_id: str,
+                 user2_slack_id: str):
+        """
+        Initialize the pairing.
+
+        Pairing ID is a UUID generated by ``uuid.uuid4()``.
```
bobheadxi, line 18:
```suggestion
        The generated ``pairing_id`` property is meant to uniquely represent a pairing.
```

something something leaky abstractions :)

app/scheduler/modules/pairing.py
```diff
@@ -0,0 +1,111 @@
...
+from config import Config
+from db.facade import DBFacade
+from app.model import Pairing
+import logging
+
+
+class PairingSchedule(ModuleBase):
+    """Module that matches 2 Launch Pad members each week"""
+
+    NAME = 'Match launch pad members randomly'
```
bobheadxi, line 17:
```suggestion
    NAME = 'Pair members randomly for meetups'
```

(pipe dream for https://github.com/ubclaunchpad/rocket2/issues/203 🙃 )

app/scheduler/modules/pairing.py
```diff
@@ -0,0 +1,111 @@
...
+from typing import Dict, List, Any, Set
+from flask import Flask
+from config import Config
+from db.facade import DBFacade
+from app.model import Pairing
+import logging
+
+
+class PairingSchedule(ModuleBase):
+    """Module that matches 2 Launch Pad members each week"""
```
bobheadxi, line 15:
```suggestion
    """Module that pairs members each ``SLACK_PAIRING_FREQUENCY``"""
```

(pipe dream for https://github.com/ubclaunchpad/rocket2/issues/203 🙃 )

app/scheduler/modules/pairing.py
```diff
@@ -0,0 +1,111 @@
...
+        users = self.bot.get_channel_users(self.channel_id)
+        logging.debug(f"users of the pairing channel are {users}")
+        matched_user_pairs = self.__pair_users(users)
+        for pair in matched_user_pairs:
+            group_name = self.bot.create_private_chat(pair)
+            logging.info(f"The name of the created group name is {group_name}")
+            self.bot.send_to_channel("Hello! \
+                You have been matched by Rocket \
+                please use this channel to get to know each other!",
+                                     group_name)
```
bobheadxi, line 46: I wonder if we should have a configuration option `SLACK_PAIRING_TIPS_LINK` that can generate a button that links to tips for what to do in a pairing (i.e. "set up video call, talk about x and y") - for Launch Pad this could be a handbook page, for example

db/dynamodb.py
```diff
@@ -183,6 +192,17 @@ def __create_table(self, table_name: str, key_type: str = 'S'):
...
+    def __enable_ttl(self, table_name: str):
+        """
+        Enable automatically dropping entries on TTL expiry
+
+        **Note**: This function should **not** be called externally, and should
+        only be called on initialization.
+
+        :param table_name: name of the table to enable ttl for
+        """
+        # Stub
```
bobheadxi, line 204:
```suggestion
        # TODO
```

(for discoverability of TODOs by searching "TODO")

Should include links to the relevant issues next to the TODOs when you create them!

interface/slack.py
bobheadxi: Ah, in that case:

```suggestion
        # We strip away the "#" in case it was provided with the channel name,
        # such as in values returned from Slack's ``conversations_list`` API.
        channel_name = channel_name.replace("#", "")
```

config/__init__.py
```diff
@@ -17,7 +17,7 @@ class Config:
         'SLACK_API_TOKEN': 'slack_api_token',
         'SLACK_NOTIFICATION_CHANNEL': 'slack_notification_channel',
         'SLACK_ANNOUNCEMENT_CHANNEL': 'slack_announcement_channel',
-
+        'SLACK_PAIRING_CHANNEL': 'slack_pairing_channel',
```
bobheadxi, line 20: Can we move this to `OPTIONALS` and make this entire feature feature-toggled on this variable?

bobheadxi (changes requested):
_Fantastic_ progress, really excited to see this land! Thanks for taking the time to do this ❤️

app/scheduler/modules/pairing.py
tarikeshaq: Alrighty made it an env variable as a cron job string... not the cleanest so happy for recommendations!

app/scheduler/modules/pairing.py
tarikeshaq: Hmmm Can we do this on a follow up or do ya think it's important? Could be a really good `first issue` for someone!

bobheadxi: Yep! Can you add all the TODO items for follow-up issues in the PR description as a checklist, to make sure we don't forget?

bobheadxi:
rocket-pairing-hd

app/scheduler/modules/pairing.py
```diff
@@ -0,0 +1,119 @@
...
+                'minute':        cron_frequency[0],
+                'hour':         cron_frequency[1],
+                'day':          cron_frequency[2],
+                'month':         cron_frequency[3],
+                'day_of_week':   cron_frequency[4],
+                'name':          self.NAME}
+
+    def do_it(self):
+        """Pair users together, and create a private chat for them"""
+        users = self.bot.get_channel_users(self.channel_id)
```
cheukyin699, line 45: Probably want to stop execution when we find out that there are no users in the channel (it also removes a check in `__pair_users()` that checks to see if there are any users in the list).

db/dynamodb.py
```diff
@@ -296,3 +316,11 @@ def delete(self, Model: Type[T], k: str):
...
             }
         )
+
+    def delete_all(self, Model: Type[T]):
+        logging.info(f"Deleting {Model.__name__}")
+        table_name = self.CONST.get_table_name(Model)
+        table = self.ddb.Table(table_name)
+        table.delete()
+        table.wait_until_not_exists()
+        self.__create_table(table_name)
```
cheukyin699, line 326: May want to comment that this is the best way to remove all items from a table, because it makes fewer calls.

interface/slack.py
```diff
@@ -111,6 +111,54 @@ def send_event_notif(self, message):
             logging.error("Webhook notif failed to send due to {} error.".
                           format(se.error))
 
+    def create_private_chat(self, users: List[str]):
+        """
+        Create a private chat with the given users
+
+        :param users: the list of users to add to the private chat
+
+        :raise SlackAPIError if the slack API returned error openning the chat
```
cheukyin699, line 120:
```suggestion
        :raise SlackAPIError if the slack API returned error opening the chat
```

sample-env
```diff
@@ -18,6 +18,7 @@ AWS_ACCESS_KEYID='53'
 AWS_SECRET_KEY='itsa secret'
 AWS_USERS_TABLE='users'
 AWS_TEAMS_TABLE='teams'
+AWS_PAIRINGS_TABLE='pairings'
```
cheukyin699, line 21: May want to add the optional config options for others to see how it works.

tests/app/model/pairing_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def setUp(self):
+        """Set up pairing variables for testing."""
+        self.p0 = Pairing('user1', 'user2')
+        self.p1 = Pairing('user1', 'user3')
+        self.p2 = Pairing('user2', 'user3')
+
+    def test_valid_pairing(self):
+        """Test the pairing static method is_valid()."""
+        self.assertTrue(Pairing.is_valid(self.p0))
```
cheukyin699, line 17: Don't forget to test the invalid cases as well!

tests/app/scheduler/modules/pairing_test.py
```diff
@@ -0,0 +1,88 @@
...
+        self.db.delete_all(Pairing)
+        self.slackbot.get_channel_users.return_value = [
+            'user1', 'user2'
+        ]
+
+        self.pairing.do_it()
+        args = self.slackbot.create_private_chat.call_args.args[0]
+        assert 'user1' in args
+        assert 'user2' in args
+        assert len(args) == 2
```
cheukyin699, line 40:
```suggestion
        self.assertIn('user1', args)
        self.assertIn('user2', args)
        self.assertEqual(len(args), 2)
```

There are others.

chuck-sys (changes requested):
Really good job on this! Seems to just need the docs now.

bobheadxi:
Nothing that this closes https://github.com/ubclaunchpad/ideas/issues/8 as well

Issue #490, opened 2020-09-25
reduce reliance on 'team refresh' (or even remove it)


bobheadxi:
I implemented #484 as a part of `team refresh`, but @cheukyin699 brought up some good points against it:

> Okay so here's the thing. /rocket team refresh takes around 20 seconds to execute (last I checked). The more members we have in Launch Pad, the longer it takes, O(n). We'll still need ways for users to receive the permission sharing automatically, whether if it is a user just adding their emails, or a team adding a folder ID into the system.
>
> I am very adamant about not using /rocket team refresh unless we really need to because ... there are lots of side effects that I don't know about, that may be disruptive. For example, today I found out that it would forcefully add a user back into the organization if the user was in ddb but was already out of the org (one possibility would be if the user ID was still in a team in ddb).

I wasn't aware of the difficulties here because I didn't delve very deeply, but this can certainly get quite concerning, and at the moment I'm not actually 100% sure #484 will work very well once rolled out to our **~200 person** Rocket deployment. Will keep an eye on this, but I'm guessing this task will encapsulate:

* [ ] decouple `all` team sync from refresh
* [ ] decouple drive permissions sync from refresh

bobheadxi:
From #504 

> I know we talked about possibly dropping refresh (#490), but we still need a way to perform the initial sync - I imagine refresh might stay around purely for use cases like this.

Updated title accordingly

PR #554, opened 2020-10-12
docs: general formatting, fix links, cleanups, rewrites

chuck-sys:
## Details

- Add links to types in internal reference
- Change the unicode directional single-/double-quotes to ascii ones so
that mac users aren't selectively favoured
- Add even more documentation on less documented functions
- Make factory documentation visible when it wasn't for so long
- Rewrite segments of architecture because it no longer reflected the
current state of the project
- Reformat some names to make internal reference links possible

View newly built docs here: https://rocket2.readthedocs.io/en/cheukyin699-update-docs/


bobheadxi:
> Change the unicode directional single-/double-quotes to ascii ones so
that mac users aren't selectively favoured

❤️

app/controller/command/commands/team.py
```diff
@@ -635,7 +639,8 @@ def refresh_helper(self, user_id) -> ResponseTuple:
         Ensure that the local team database is the same as GitHub's.
 
         In the event that our local team database is outdated compared to
-        the teams on GitHub, this command can be called to fix things.
+        the teams on GitHub, this command can be called to fix these
+        inconsistencies.
```
bobheadxi, line 643: This likely ties into https://github.com/ubclaunchpad/rocket2/issues/490 in that we can't remove this command, as we'll always need something to forcibly bring everything in line

conf.py
```diff
@@ -23,9 +23,9 @@
 author = 'UBC Launch Pad'
 
 # The short X.Y version
-version = '1.0'
+version = '2.0'
 # The full version, including alpha/beta/rc tags
-release = '1.0.0'
+release = '2.0.0'
```
bobheadxi, line 28: do we even do releases?

config/__init__.py
```diff
@@ -118,7 +118,7 @@ def __init__(self, missing_config_fields):
         """
         Initialize a new MissingConfigError.
 
-        :param: missing_config_fields List of missing config variables
+        :param missing_config_fields: the missing config variables
```
bobheadxi, line 121: is there a linter that can catch these documentation errors?

docs/Architecture.rst
```diff
@@ -3,18 +3,22 @@ Architecture
...
 
-**What is the ``command`` module?** The ``command`` module is where the
-slack commands get parsed and passed on to the backend so models can be
-created and the database can be populated.
+We store our data in an Amazon DynamoDB, which can be accessed directly by the
+database facade :py:class:`db.dynamodb.DynamoDB`.
 
-**What is the ``model`` module?** The ``model`` module is where models
-are constructed. Currently we have ``Team`` and ``User`` models.
+We treat Github itself as the sole source of truth. Any modifications done on the
```
bobheadxi, line 14:
```suggestion
We treat GitHub itself as the sole source of truth. Any modifications done on the
```

etc - though we can also have a separate PR to make all these fixes in one go

Reference: official branding

image

bobheadxi:
This might be because, to center-align the "Rocket 2" header in both GitHub README and the website, I use HTML, which isn't recognized by Sphinx as the top-level header for the page - is there a way to override this? (should be "Rocket 2", not "Usage")

image

config/__init__.py
cheukyin699: I doubt it.

conf.py
cheukyin699: Good question. Actually, there is a slim chance that this would come back and bite us. Technically we do releases. We are currently on v2.1.0. I'm thinking that'd be one of the things we'd talk about in the meeting?

Gonna change this to better reflect, but right now docs are generated (and not saved) on master push, soooo rolling release.

app/controller/command/commands/team.py
cheukyin699: Probably need to add a `Deprecation Warning` whenever someone calls the command.

conf.py
bobheadxi: I dropped an item in the meeting notes - feel free to add stuff! https://docs.google.com/document/d/1KcH2l6nJKyHoCDe6v6Sl0cLWQWKQQmSLqk_kXvZUsRw/edit

app/controller/command/commands/team.py
bobheadxi: Oh my point was I _dont_ think we can deprecate it, ever, since we need some way to force a state sync (for example, to add new features - essentially a "migration")

cheukyin699:
> Is there a way to override this?

I just noticed. Good question. After a bit of read, I think it's safe to say that, yes, there is a way to override this. And **no**, it's not pretty. See 62a6e82.

bobheadxi:
@cheukyin699 sadly, your approach in 62a6e82f517e84fd87880f085db10c36b0980485 doesn't work on GitHub :( that's why I made the entire top section raw HTML previously - so that it would work on GitHub as well (IMO this is a requirement)

image

cheukyin699:
@rob ahhh. Ohhhh. Welp, back to the drawing board. Might not be possible then.

bobheadxi:
Let's leave that for another PR - can you revert 62a6e82 and I'll approve?

cheukyin699:
Another option:

```rst
.. raw:: html

	<style>
		h1:nth-of-type(2) {
			display: none;
		}
	<style>
```

But that would be way too hacky for me.

cheukyin699:
@bobheadxi reverted

bobheadxi (approved):
thank you!!!!

bobheadxi:
Created https://github.com/ubclaunchpad/rocket2/issues/555 as follow-up

Issue #560, opened 2020-10-15
proposal: switch PR merge strategy to squashing


bobheadxi:
## Summary

Switch the Rocket 2 project's pull request merge strategy to squashing.

## Requirements

The GitHub repository is updated to only accept squash merges when landing pull requests, and docs are updated to reflect that (they currently explicitly request that we do not squash, to preserve history).

## Context

With 1553 commits at time of writing, commits are essentially just noise at this point - lacking in context, overly numerous, and more often than not reflecting WIP states that do not reflect the actual state of changes in the repository (ie the goal is to preserve history, but it's not actually history)

Using a [commit-squashing strategy](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits) explicitly ties each change to a PR and the associated discussions, which:

* improves `git blame` - you can actually see the full context of the change
* incentivize better commit bodies, since you are prompted to provide one at merge time
* less need to worry about beautiful commits when work is WIP, and more freedom to provide lots of commits
* encourages smaller-scoped PRs, since it becomes confusing to provide multiple changes in a single PR if they must end up squashed

The following Launch Pad projects also use squash:

* https://github.com/ubclaunchpad/inertia - switched to squash at ~1000 commits
* https://github.com/ubclaunchpad/docs
* https://github.com/ubclaunchpad/ubclaunchpad.com

bobheadxi:
cc @cheukyin699 for thoughts

cheukyin699:
I like it, though it will take us longer to get to the next goal of 2000 commits.

bobheadxi:
Just send 2000 pull requests - easy cheese

cheukyin699:
Something something hacktoberfest

cheukyin699:
I'll set it up now?

bobheadxi:
Yep! Just change it over and update the docs for it to close out this issue :) Thank you!

Issue #510, opened 2020-09-29
drive invites seem to be spammed for a few people


bobheadxi:
I have not experienced this myself, and judging from the general lack of complaints this is generally not the case, but there is at least once case: https://ubclaunchpad.slack.com/archives/CK93HTYQN/p1601316108020500

* @jedyeo
* @ctong25

I'm surprised this is happening, because the code excludes existing shares:

https://sourcegraph.com/github.com/ubclaunchpad/rocket2/-/blob/interface/gcp.py#L38-49

We do log the entire permissions object, so we might be able to check the logs to see if there are any without an `emailAddress`. The email invite might contain the permission ID for easier search across the logs.

bobheadxi:
I suspect this is caused by https://github.com/ubclaunchpad/rocket2/issues/497, and _might_ be partially resolved by #504 with 35e1209

jedyeo:
![image](https://user-images.githubusercontent.com/33077039/94600922-860f1100-0247-11eb-8e3e-32fcddbf173a.png)

Rocket keeps spamming me with invites every couple of hours, so each time I accept the invitation, I gain permissions for the drive, but the next invite also seems to revoke those privileges.

This only seems to be happening for some users.

Timestamps of the invites:
Sept-27 3:12pm
Sept-27 6:31pm
Sept-27 8:36pm
Sept-28 12:56am
Sept-28 3:17am
Sept-28 8:52am
Sept-28 6:39pm

The invites seem to have stopped for now. Unfortunately I do not see anywhere on the email that contains a permission ID, so what I did was accept the invitation and was able to capture something that resembled a UID at the end of the browser link: `5f729053`. Sadly I do not know if this corresponds to anything.

@bobheadxi

jedyeo:
OK, I just got another invite at 12:07pm today

ctong25:
I also got it at that time too.
Best regards,
Claudia Tong
tong.claudia.25@gmail.com

bobheadxi:
Thanks everyone!

@cheukyin699 do you mind checking if there's any permissions entries logs at some of these times where the permissions don't contain an `emailAddress`? (or DM me with how I can get myself set up with logs access)

bobheadxi:
Changes from #504 did not seem to alleviate this issue. I'm almost certain it is caused by #497, but do not understand why it's not affecting some people (ie I don't receive duplicate invites)

bobheadxi:
[...]

PR #563, opened 2020-10-16
gcp: respect permissions inherited from parent drives

### LINKED ISSUE #497: drive: permission invites get re-sent on occasion

bobheadxi:
Drive shares frequently get recreated because we fail to page through all results (we likely have over 100 shares for large folders), and when deletion was enabled (prior to https://github.com/ubclaunchpad/rocket2/pull/540) we would also erroneously delete inherited permissions. Both of these issues result in many users receiving duplicate invites whenever Rocket performs a sync (#510, #497). To amend this, we make the following changes:

* Permissions listing is now paginated
* Permissions listing now happens for direct parents of target folders - permissions discovered in parents are treated as "inherited", and for these permissions we skip both recreation and deletion.

## Ticket(s)

Closes (hopefully) #497 

## Details

This was really gnarly.

Note that this looks quite convoluted because there is no way in Drive to detect if a permission is inherited (attempt was made in https://github.com/ubclaunchpad/rocket2/pull/499) - that functionality is restricted to "Shared Drives", which we do not pay for. There are most misc. comments and context in #497 and #510 

I have also not performed an integration test because I'm inclined to just yolo-deploy and revert if broken, which should be easy to do now that we've adopted #560 🙃


interface/gcp.py
```diff
@@ -17,54 +20,110 @@ def __init__(self, drive_client: Resource, subject=None):
...
+    def ensure_drive_permissions(self,
+                                 team_name: str,
+                                 drive_id: str,
+                                 emails: List[str]):
         """
-        Create permissions for the given emails, and removes everyone not on
-        the list.
+        Create permissions for the given emails on the given Drive item, and
+        removes everyone not on the list, to ensure the state of shares on the
+        Drive item matches the emai list provided.
```
cheukyin699, line 82:
```suggestion
        Drive item matches the email list provided.
```

interface/gcp.py
```diff
@@ -17,54 +20,110 @@ def __init__(self, drive_client: Resource, subject=None):
...
+            parents: List[str] = file['parents']
+            if len(parents) > 0:
+                return parents
+        return []
+
+    def get_drive_permissions(self, drive_id: str) -> List[GCPDrivePermission]:
+        """
+        Retrieves list of permissions present on the given Drive item. It
+        pages through all results and returns a subset of permission fields,
+        as defined in `GCPPermission`.
```
cheukyin699, line 42:
```suggestion
        as defined in :class:`GCPDrivePermission`.
```

interface/gcp.py
```diff
@@ -17,54 +20,110 @@ def __init__(self, drive_client: Resource, subject=None):
...
+            .list(fileId=drive_id,
+                  fields=", ".join(fields),
+                  pageSize=50)
+        while next_page_req is not None:
+            list_res = next_page_req.execute()
+            permissions: List[Any] = list_res['permissions']
+            for p in permissions:
+                if 'emailAddress' in p:
+                    perm = GCPDrivePermission(p['id'], p['emailAddress'])
+                    perms.append(perm)
```
cheukyin699, line 62: Abstract pagination away with Python generators?

interface/gcp.py
```diff
@@ -17,54 +20,110 @@ def __init__(self, drive_client: Resource, subject=None):
...
-        the list.
+        Create permissions for the given emails on the given Drive item, and
+        removes everyone not on the list, to ensure the state of shares on the
+        Drive item matches the emai list provided.
 
-        In all cases of API errors, we log and continue, to try and get close
-        to the desired state of permissions.
+        It respects permissions inherited by one level of parents of the given
+        Drive item - permissions inherited from two levels of parents are at
+        risk of being deleted it the user is not on the provided email list.
```
cheukyin699, line 86:
```suggestion
        risk of being deleted if the user is not on the provided email list.
```

interface/gcp.py
```diff
@@ -17,54 +20,110 @@ def __init__(self, drive_client: Resource, subject=None):
...
+        inherited: List[str] = []  # emails
+        try:
+            parents = self.get_drive_parents(drive_id)
+            for parent_id in parents:
+                perms = self.get_drive_permissions(parent_id)
+                for p in perms:
+                    inherited.append(p.email)
+        except Exception as e:
+            logging.warn("Unable to fetch parents for drive item"
+                         + f"({team_name}, {drive_id}): {e}")
```
cheukyin699, line 106: Put into another function please

interface/gcp.py
```diff
@@ -86,35 +146,30 @@ def set_drive_permissions(self,
...
+                    .delete(fileId=drive_id,
+                            permissionId=p_id)\
+                    .execute()
+                deleted_shares += 1
+            except Exception as e:
+                logging.error(
+                    f'Failed to delete permission {p_id} for '
+                    + f'drive item ({team_name}, {drive_id}): {e}')
+        logging.info(
+            f'Deleted {deleted_shares} permissions for {team_name}')
```
cheukyin699, line 172: Put into another function please.

chuck-sys (changes requested):

interface/gcp.py
bobheadxi: I'd rather keep this one here, since it is tightly coupled with the permissions creation - I don't want to expose an API that does anything other than `ensure` a permissions state (ie no create-only or delete-only APIs)

interface/gcp.py
bobheadxi: 58cfe4ce903af8382d3ddec91778f43dd297555d

interface/gcp.py
bobheadxi: Eh took a look at it and seems unnecessary and convoluted? Happy to take a suggestion if you believe theres a way to make this more succinct by setting up a generator for this single request (none of the others used, including parent listing, requires paging)

cheukyin699: Oh. It just makes it easier for me to read. 

```python
def get_paginated_permissions(self, drive_id: str, fields: List[str]): Iterator[Any]:
	req = self.drive.permissions()\
		.list(fileId=drive_id,
			  fields=', '.join(fields),
			  pageSize=50)
	while req is not None:
		resp = req.execute()
		for perm in resp['permissions']:
			yield perm
		req = self.drive.permissions().list_next(req, resp)
	raise StopIteration
```

cheukyin699: I tried this, but then the tests started failing. Help.

```diff
diff --git a/interface/gcp.py b/interface/gcp.py
index 78830b5..e6a29da 100644
--- a/interface/gcp.py
+++ b/interface/gcp.py
@@ -1,5 +1,5 @@
 """Utility classes for interacting with Google APIs"""
-from typing import Any, List
+from typing import Any, List, Iterator
 from googleapiclient.discovery import Resource
 import logging
 
@@ -35,6 +35,27 @@ class GCPInterface:
                 return parents
         return []
 
+    def get_paginated_permissions(self,
+                                  drive_id: str,
+                                  fields: List[str]) -> Iterator[Any]:
+        """
+        Retrieve generator of permissions present on the given Drive item.
+
+        Pages through them so you don't have to.
+        """
+        req = self.drive.permissions()\
+            .list(fileId=drive_id,
+                  fields=', '.join(fields),
+                  pageSize=50)
+
+        while req is not None:
+            resp = req.execute()
+            for perm in resp['permissions']:
+                yield perm
+            req = self.drive.permissions().list_next(req, resp)
+
+        raise StopIteration
+
     def get_drive_permissions(self, drive_id: str) -> List[GCPDrivePermission]:
         """
         Retrieves list of permissions present on the given Drive item. It
@@ -42,34 +63,18 @@ class GCPInterface:
         as defined in :class:`GCPDrivePermission`.
         """
         perms: List[GCPDrivePermission] = []
-        page_count = 0
         fields = [
             "permissions/id",
             "permissions/emailAddress",
         ]
         # See http://googleapis.github.io/google-api-python-client/docs/dyn/drive_v3.permissions.html#list # noqa
         # pylint: disable=no-member
-        next_page_req = self.drive.permissions()\
-            .list(fileId=drive_id,
-                  fields=", ".join(fields),
-                  pageSize=50)
-        while next_page_req is not None:
-            list_res = next_page_req.execute()
-            permissions: List[Any] = list_res['permissions']
-            for p in permissions:
-                if 'emailAddress' in p:
-                    perm = GCPDrivePermission(p['id'], p['emailAddress'])
-                    perms.append(perm)
-
-            # set state for the next page
-            # see https://googleapis.github.io/google-api-python-client/docs/dyn/drive_v3.permissions.html#list_next # noqa
-            # pylint: disable=no-member
-            next_page_req = self.drive.permissions()\
-                .list_next(next_page_req, list_res)
-            page_count += 1
-
-        logging.info(f"Found {len(perms)} permissions across {page_count} "
-                     + f"pages for {drive_id}")
+        for p in self.get_paginated_permissions(drive_id, fields):
+            if 'emailAddress' in p:
+                perm = GCPDrivePermission(p['id'], p['emailAddress'])
+                perms.append(perm)
+
+        logging.info(f"Found {len(perms)} permissions for {drive_id}")
         return perms
 
     def get_parents_permissions(self,

```

bobheadxi:
```
---------------------------------------------------------------- Captured log call -----------------------------------------------------------------
WARNING  root:gcp.py:118 Unable to fetch parents for drive item(team, target-drive): name 'Iterator' is not defined
ERROR    root:gcp.py:141 Failed to load permissions for drive item(team, target-drive): name 'Iterator' is not defined
```

Working on applying a version of your patch

bobheadxi: 1df27bbb12f673fea18436c797373709f8ee2536

I'm against having the paginated function be a to-level API for a similar reason to https://github.com/ubclaunchpad/rocket2/pull/563#discussion_r506576359 (don't expose unused APIs), so I scoped it to the only function that will use it (this one)

chuck-sys (approved):

PR #577, opened 2020-10-17
deps: set stricter dependency requirements

bobheadxi:
This will help Dependabot out, and also make `pipenv install` more consistent for everyone


cheukyin699:
These might be updates, and judging from the new linting errors, it seems that our favourite slack api library decided that it was high time to add typing, meaning that `mypy` is correct. Needs to do some casting and whatnot.

bobheadxi:
@cheukyin699 do you mind taking over this PR?

cheukyin699:
@bobheadxi I have fellowship for the next 2 hours so I'll hopefully get to it later?

Issue #543, opened 2020-10-06
Remove `200` codes from ResponseTuples in function signatures


cheukyin699:
## Summary

Remove `200` codes from ResponseTuples in function signatures

## Requirements

All commands, whether they do something well or not (unless in the case of internal server failure, in which case a status code of `5xx` is auto-returned), *must* return `200`, [no matter what](https://api.slack.com/interactivity/slash-commands#responding_with_errors). If that's the case, we should return 200 for all of them and just pull it up into the handle function.

PR #583, opened 2020-10-29
tests: increase clarity by a marginal amount, **rm api and tests/api**

### LINKED ISSUE #531: Testing isn't what it seems to be
Update current command tests (focusing on team and user commands as a priority) and refactor to make it look cleaner.

As of currently, it is a hell-scape. Mocks are strewn here and there, and they are really hard to read. Make it easier for maintainers (like me) to read. Use the more modern ways of doing things (like helper utility functions and helper functions in general and a fully-utilized memdb.

Pretty please?

chuck-sys:
## Ticket(s)

Closes #531

## Details

Since I am reasonably satisfied with the size of the tests (for some reason even after I made the PR from last time this still stays), I will close this. Should have closed this like 23 days ago.

- rename deep-dive to user inspect because we are using `--inspect`
- remove most `self.assertEqual(code, 200)` because there is no other possible code that you can actually output (so no testing)
	- however, all of the `self.assertTupleEqual()` functions weren't touched because removing the status code would make it more difficult to read
	- planning on completely removing when we get through #543 
- **remove `api` and `tests/api` to alleviate confusion on inconsistency between command logics**


bobheadxi (approved):
im not entirely sure what happened here, but i trust it's a good decision, and similar coverage using less tests seems alright :)

PR #594, opened 2020-11-08
feat: add ability to change github team name

### LINKED ISSUE #556: team edit: add ability to change github team name
## Summary

Add the ability to run `/rocket team edit --github []`, and rename `--name` to `--displayname`

## Requirements

We need to be able to rename GitHub teams from Rocket.

## Context

Lots of teams change their names after creation, but are unable to change their team names

no-or:
##556


cheukyin699:
See #270. `argparse` auto-converts arguments that contain underscores to hyphens, so you don't need to use underscores.

PR #606, opened 2020-12-04
feat: add ability to change github team name

no-or:
##556


chuck-sys (changes requested):
Okay, so when we said we wanted the ability to change Github team names, we meant to have the ability to change the names not just on Rocket, but also on Github itself. You are gonna need to hook this up to Github API.

no-or:
@cheukyin699 Hey, I thought about what you said. The context of the ticket says, "Lots of teams change their names after creation, but are unable to change their team names" which kinda directed me to change the name only on slack. But, I can look into it the Github API if needed no worries :))

cheukyin699:
> @cheukyin699 Hey, I thought about what you said. The context of the ticket says, "Lots of teams change their names after creation, but are unable to change their team names" which kinda directed me to change the name only on slack. But, I can look into the Github API if needed no worries :)) Please confirm that Github needs to be hooked up. Thanks!

Yes. Github needs to be hooked up.

bobheadxi:
> The context of the ticket says, "Lots of teams change their names after creation, but are unable to change their team names" which kinda directed me to change the name only on slack

Cheuk's right here, from #556 : 

> We need to be able to rename GitHub teams from Rocket.

The point of this is that we need to be able to change team names such that it actually changes things - ie the new name propagates across connected platforms (namely GitHub)

chuck-sys (dismissed):

